PR #63, opened 2018-02-01
Close #60: address resolution for gateway

### LINKED ISSUE #60: Allow the gateway to connect to resolvable addresses  instead of only ip-addresses
This would also allow the bootstrap nodes to be in DNS and make docker setups easier

LeeSmet:


LeeSmet:
After giving this some thought, I moved the DNS resolution calls to the api endpoints (gateway connect and gateway disconnect). This ensures that the internal code always has a NetAddress with IP. This is similar behavior to the bootstrap nodes, which are converted the first time they are iterated. If any endpoints are added which take a NetAddress parameter in the future, a call must then be added to convert the address, but imo the single call/line of code required is not really any notable overhead

LeeSmet:
- TryDNSResolution renamed to TryNameResolution
- Docker script has been modified to create a new docker network (to take advantage of docker's build in name resolution)
- Dockers are now started in the new network, and connected via both the *container_name:port* and *container_name.network_name:port* to demo this ability

PR #73, opened 2018-02-08
Fix blockcreator race

LeeSmet:
Solve a reasonably frequent race condition when the block creator tries to solve a block just as it receives a consensus change. Also avoid creating blocks until we are fully synced to save some  system resources that would otherwise be wasted on useless blocks, especially during IBD on a new node (probably nothing major).

I feel that it is better to cache the synced state on the blockcreator struct, as this avoids a (relatively) expensive readlock to check the consensus sync state itself. Also, it only ever switches from false to true once we caught up after a (re)start of rivined.

As for the readlock on the `solveBlock` metod, I'm not sure if it is best to readlock here just to lock the height variable from the consensus change, or if we should lock in the parent method (`SolveBlocks`), locking untill we submitted a possible block to the consensus set.

### DIFF
--- a/modules/blockcreator/blockcreator.go
+++ b/modules/blockcreator/blockcreator.go
@@ -19,6 +19,9 @@ type BlockCreator struct {
 	tpool  modules.TransactionPool
 	wallet modules.Wallet
 
+	// Cache the synced state of the consensus set to avoid unnecessarily locking it
+	csSynced bool
+
 	unsolvedBlock *types.Block
 
 	log        *persist.Logger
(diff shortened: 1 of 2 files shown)


LeeSmet:
The only reason to switch sync back to false is if we have some kind of network failure where a synced node is disconnected for a considerable amount of time. Though I don't think any special kind of handling is currently defined for this case. Behaviour could be easily tested though if we have some kind of test setup where we can bring down the network of one node for a while and observe what it does when it comes back online

LeeSmet:
Also, even we switch the synced flag back to false in the consensusset, this would mean that the blockcreator would try to create some useless blocks where it otherwise does nothing, so it wouldn't be the end of the world

robvanmieghem:
It slows down the syncing a bit if a node has to catch up now but I would lower the  8 seconds wait if we would reset synced flag

PR #88, opened 2018-02-12
Release container images

### LINKED ISSUE #76: integrate the publish script in the makefile as a publish target
and add zero-os container image creation

LeeSmet:
Closes #76 

- Change the xc Makefile recipe to use a docker builder
- Build a minimal docker from the linux binaries, version is taken from git describe and trimmed to a `vX.X.X` format
- Push this image to dockerhub, in the rivine repo
- Convert this image from dockerhub to a zero-os flist on the gig hub

This requires:
- To be logged in on dockerhub, with the ability to push images to rivine
- A valid jwt from itsyouonline exported in the `HUB_JWT` key. This jwt requires a `user:memberof:rivine scope`

example of creating such a JWT: 

```
export HUB_JWT=$(curl -X POST "https://itsyou.online/v1/oauth/access_token?response_type=id_token&grant_type=client_credentials&client_id=rBHE0wxxx...&client_secret=mVxfCeqVZtLcxxx...&scope=user:memberof:rivine")

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -7,6 +7,7 @@ all: install
 run = Test
 pkgs = ./build ./modules/gateway ./rivined ./rivinec
 testpkgs = ./types ./build ./crypto ./encoding ./modules ./persist ./rivinec ./rivined ./sync ./types
+version = $(shell git describe | cut -d '-' -f 1,3 |  cut -d '-' -f 1)
```
GlenDC: Can be simplified to `version = $(shell git describe | cut -d '-' -f 1)`.

Dockerfile
```diff
@@ -2,9 +2,8 @@ FROM golang:1.6.3
 MAINTAINER rivine.io
 
 ENV CGO_ENABLED 0
-COPY . /go/src/github.com/rivine/rivine
 WORKDIR /go/src/github.com/rivine/rivine
 
-EXPOSE 23112
```
GlenDC: We don't need this expose any longer?

Dockerfile
```diff
@@ -2,9 +2,8 @@ FROM golang:1.6.3
 MAINTAINER rivine.io
 
 ENV CGO_ENABLED 0
-COPY . /go/src/github.com/rivine/rivine
```
GlenDC: No need to copy any longer?

Makefile
```diff
@@ -33,7 +34,14 @@ release-std:
 # for all windows, linux and mac, 64-bit only,
 # using the standard Golang toolchain.
 xc:
-	./release.sh
+	docker build -t rivinebuilder .
+	docker run --rm -v $(shell pwd):/go/src/github.com/rivine/rivine rivinebuilder
```
GlenDC: Why not `$GOPATH/src/github.com/rivine/rivine`?

Makefile
```diff
@@ -33,7 +34,14 @@ release-std:
...
+
+# Release images builds and packages release binaries, and uses the linux based binary to create a minimal docker
+release-images: xc
```
GlenDC: it's a weird target name, as we only release 1 image?

Perhaps we could call this `release-docker` instead?

Makefile
```diff
@@ -33,7 +34,14 @@ release-std:
...
+	docker build -t rivine/rivine:$(version) -f DockerfileMinimal --build-arg binaries_location=release/rivine-$(version)-linux-amd64 .
+	docker push rivine/rivine:$(version)
+	curl -b "active-user=rivine; caddyoauth=$(HUB_JWT)" -X POST --data "image=rivine/rivine:$(version)" "https://hub.gig.tech/api/flist/me/docker"
```
GlenDC: Perhaps we could fail the Make command when `HUB_JWT` is not defined? Might provide better UX than whatever vague error you'll get otherwise.

Dockerfile
```diff
@@ -2,9 +2,8 @@ FROM golang:1.6.3
...
 WORKDIR /go/src/github.com/rivine/rivine
 
-EXPOSE 23112
+RUN apt-get update && apt-get install -y zip
 
-ENTRYPOINT go install -v -tags 'debug dev' ./... && rivined --no-bootstrap
+ENTRYPOINT ./release.sh
```
GlenDC: Please adapt the `release.sh` script, such that we can ensure to only build one OS (e.g. Linux), rather than all of them, because it doesn't make sense to bio;d all of them, to only use one.

Perhaps by default we assume you want all targets (OSs), but you can also specify one or multiple OSs as optional positional arguments (e.g. `./release.sh linux` or `./release.sh darwin linux`), e.g.

By the way we might also want to provide a FreeBSD build in `./release.sh`.

Makefile
GlenDC: I also don't think this should rely on `xc`, as we in the end only care for this one about Linux, or am I missing something here?

GlenDC:
> To be logged in on dockerhub, with the ability to push images to rivine

Would the docker commands not ask to be logged in, should you not be logged in yet?

> example of creating such a JWT

I think it could be nice if this could be done automatically, either as part of that target or as a separate script or Make target, as that is something that will have to be done pretty much each time you push to docker, given the expiration time of such a token.

Dockerfile
LeeSmet: This docker is only used for the build proces, it no longer runs the binary

Dockerfile
LeeSmet: Copy would invalidate the docker cache if any of the copied files changes/gets deleted/a new one is added. By mounting the entire directory later the entire docker image is kept in cache (untill a change is made in this file to modify the build environment)

Makefile
LeeSmet: I feel that its highly unlikely that someone will invoke make from a different directory, so this should be identical. Then again, the presence of $Gopath usually relies on a golang installation on the local machine, which is otherwise not required as the compilation is done in the docker. Although this also seems unlikely, I prefer this approach

Makefile
LeeSmet: We release 2: a docker container and a zero-os flist. Therefore  I feel like `release-docker` would also not be a suitable name. While its true that we only use linux binaries for the containers, in practice this will always be done as part of the release process. So by having it depend on `xc` the entire release process only requires one invocation of `make release-images` instead of `make xc; make release-images`

Makefile
LeeSmet: Then again, the JWT might also be outdated, or not contain the `user:memberof:rivine` scope. Since releasing is usually done by a developer, I think we can assume at least **some** knowledge on his part, and not require extensive hand-holding as is the case with end users.

When talking about UX, do we really want to check if the env var is set and return an error if its not, to then not validate whether this JWT is valid or not?

LeeSmet:
>>To be logged in on dockerhub, with the ability to push images to rivine

> Would the docker commands not ask to be logged in, should you not be logged in yet?

To be honest, I don't really know. I did it separately half a year ago and I'm still logged in on the cli. If it automatically asks for login credentials, all the better, should only occur once unless you switch to a new dev machine.

The main takeaway here is actually: Your docker hub account needs to be able to push to rivine on dockerhub

>>    example of creating such a JWT

>I think it could be nice if this could be done automatically, either as part of that target or as a separate script or Make target, as that is something that will have to be done pretty much each time you push to docker, given the expiration time of such a token.

Creating a new one from a client_id and client_secret would require those to be specified somewhere. It obviously can't be in the Makefile, or a file which isn't gitignored, so it requires manual configuration (albeit only once). A more suitable option would be to require the JWT set in the env var to be refresheable and call refresh for every push imo, but this has the limitation that a refresh token is only valid for 30 days since the last time it was used.

Dockerfile
LeeSmet: Since the script is only used as part of the release flow, I'm not sure why we'd ever not want to build **all** supported targets?

Also do we want to support release zips for different OS's now, or do we wait until someone actually informs about them... There are a lot of different OS's like this we could add as go supports them, but for now I think they would just slow down build process without much benefit.

LeeSmet:
Fixed the version string

GlenDC:
> Creating a new one from a client_id and client_secret would require those to be specified somewhere.

Defining client_id and client_secret env variables is a lot easier (and defining env variables as part of build processes is a pretty common thing to require anyhow), compared to having to remember an entire REST API, and having to define there the client_id and client_secret anyhow.

Dockerfile
robvanmieghem: this only works wit a local docker( or local vm)

Dockerfile
robvanmieghem: indeed, but the doicker was also there  to just run the thing, now that it's only a builder, the "Dockerfile" should be renamed  to make it clear that it only serves 1 purpose

Dockerfile
GlenDC: Can you explain further @robvanmieghem, I'm afraid you lost me.

Makefile
GlenDC: Fair enough, it's a pretty flaky release pipeline none the less, but for now it's OK, we can come back later on this. Really not that important for now.

I do start to feel it would be useful if IYO provides a command-line client, which would provide the API as commands, as well as provide ssh access (if possible), as the latter would really make it not necessary to provide any credentials at all.

GlenDC:
Also make sure to rebase master into your branch, as you're out-of-date.

LeeSmet:
- Update `Dockerfile` to `DockerBuilder`
- Auto generate JWT when you release the hub image:
  - This now requires a `HUB_APP_ID` and `HUB_APP_SECRET` to be set
  - Check if these are set before trying to get the JWT

DockerBuilder
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.6.3
```
GlenDC: Shouldn't we aim for a higher version? Rivine is new software, so it's a bit weird to keep maintaining for an old version of Golang. I would say we can easily go for 1.9.x as our minimum version, especially as 1.10.x has been released yesterday already.

Makefile
```diff
@@ -83,4 +91,13 @@ update_deps:
 	dep ensure -v
 	dep ensure -update -v
 
+get_hub_jwt: check-HUB_APP_ID check-HUB_APP_SECRET
+	$(eval HUB_JWT = $(shell curl -X POST "https://itsyou.online/v1/oauth/access_token?response_type=id_token&grant_type=client_credentials&client_id=$(HUB_APP_ID)&client_secret=$(HUB_APP_SECRET)&scope=user:memberof:rivine"))
+
+check-%:
+	@ if [ "${${*}}" = "" ]; then \
+		echo "Env var $* not present"; \
```
GlenDC: I would change this to `Required Env var`

Makefile
```diff
@@ -33,7 +34,14 @@ release-std:
 # for all windows, linux and mac, 64-bit only,
 # using the standard Golang toolchain.
 xc:
-	./release.sh
+	docker build -t rivinebuilder -f DockerBuilder .
+	docker run --rm -v $(shell pwd):/go/src/github.com/rivine/rivine rivinebuilder
+
+# Release images builds and packages release binaries, and uses the linux based binary to create a minimal docker
+release-images: make_hub_jwt xc
```
GlenDC: shouldn't this be `get_hub_jwt`?

GlenDC:
When running `make xc`, building it works, but the run step fails:

```
docker run --rm -v /Users/glendc/go/src/github.com/rivine/rivine:/go/src/github.com/rivine/rivine rivinebuilder
/bin/sh: 1: ./release.sh: not found
make: *** [xc] Error 127
```

LeeSmet:
Bumped the go version to 1.9.4 while building, modified the error msg, and fixed the target in the makefile

Not sure why you are getting errors, `xc` builds fine for me.

GlenDC:
> Not sure why you are getting errors, xc builds fine for me.

Fairly sure because you no longer copy the workdir first (`COPY . /go/src/github.com/rivine/rivine`).   No where in your `DockerBuilder` file do you specify to get the files into the docker container. It's just a pure Golang container, as far as I can see.

LeeSmet:
That doesnt really explain why it works for me. Also, after updating the golang version to 1.9.4., the entire builder got recreated, so it can't be some magical caching issue. The files are indeed not copied, the workdir is just set and then the building directory is mounted there. Finally, the docker image is able to use apt-get, so that would indicate that its based on (some) debian (derivative) (not sure exactly which one though)

GlenDC:
> Finally, the docker image is able to use apt-get, so that would indicate that its based on (some) debian (derivative) (not sure exactly which one though)

Yes, because the Golang container is based on a debian one.

If we start from Go 1.94:

https://github.com/docker-library/golang/blob/366fe83ed839938cd04b2d546a06e2aee25a39a2/1.9/stretch/Dockerfile

We see it's based on...

+ https://github.com/docker-library/buildpack-deps/blob/1845b3f918f69b4c97912b0d4d68a5658458e84f/stretch/scm/Dockerfile

Which itself is based on...

https://github.com/docker-library/buildpack-deps/blob/d662dc69f910feb241f6d0c9d2cd6cc2fb6c5e6c/stretch/curl/Dockerfile

Which is finally based on `Debian:stretch`: https://github.com/debuerreotype/docker-debian-artifacts/blob/132a85df5e5e1528b46bcd44e8bfcc9d82ffce2d/stretch/Dockerfile

GlenDC:
> That doesnt really explain why it works for me. Also, after updating the golang version to 1.9.4., the entire builder got recreated, so it can't be some magical caching issue. The files are indeed not copied, the workdir is just set and then the building directory is mounted there

Sounds good, I reset my docker environment to factory settings and now it does work fine. I guess i screwed something up during a previous experiment.

GlenDC:
Works fine now, made some last improvements, and than it is fine I think.

GlenDC:
In the future we could improve the pipeline even further by also automatically making a release and uploading those zip files to GitHub, which should all be possible using the GitHub API I can imagine, which seems like a logical step after having already the upload to the GIG Hub.

But for now I think this is OK as it is. It's not like we release that often anyhow, and making a release on GitHub is easy enough as it is.

PR #127, opened 2018-03-06
add a registerdata command

### LINKED ISSUE #117: Add a "registerData" command to the daemon and client
A transaction is created that puts the data in the arbitrary data field and pays the transaction fee,, a wallet needs to be unlocked for this to work off course.
Documentation on how to do this through the api is required as well

LeeSmet:
There is no check on the size of the arbitrary data since I could not find a reference to a maximum size (thought there was one). If the data is too large, then the transaction will be rejected by the transaction pool anyway so this isn't an issue I think.

Closes #117


GlenDC:
I think it is OK for now as it is.

PR #134, opened 2018-03-07
Close #128: Multiple networks in the same binary

### LINKED ISSUE #128: allow for different networks of the same type of blockchain
We need to allow different networks using the same type of blockchain (that is to say the same (custom) rivine implementation), to connect to different networks, without mixing connections. For example if we have a prod and test net, we don't want to by accident create test blocks on the prod network.

Therefore we'll need to be able to configure the initial parameters (e.g. genesis related configs), as well as bootstrap nodes. This needs to be possible using a config file, but it should also be possible to have them embedded already. (e.g. it makes sense that the default is the production one, and that you could allow for a flag to the daemon as to allow to connect to the default test network instead).

This feature is required for tfchain, as can be seen in https://github.com/threefoldfoundation/tfchain/issues/3

LeeSmet:
Some of the types variables are still initialized instead of only declared since the tests don't seem to like it otherwise


GlenDC:
Will review later, however I can already tell you that you will have to rebase our branch and resolve some conflicts, as the rivinec/rivined packages have been renamed and moved.

LeeSmet:
weird since I pulled master and created my branch just before I committed these changed and pushed them

GlenDC:
> weird since I pulled master and created my branch just before I committed these changed and pushed them

It's because I pushed those changes when you just made the PR, I saw that too late. Either way, pushed those changes as early as possible, as I didn't want them to impact all those config-related features we'll implement over the next couple of days.

types/constants.go
```diff
@@ -50,200 +50,153 @@ var (
...
 }
 
-// CalculateGenesis fills in the genesis block variables which are computed based on
-// variables that have been set earlier
-func CalculateGenesis() {
+// DefaultChainConstants provide sane defaults for a new chain. Not all constants
+// are set, since some (e.g. GenesisTimestamp) are chain specific, and this also
+// allows some santiy checking later
+func DefaultChainConstants() ChainConstants {
+	// GenesisTimestamp, GenesisBlockStakeAllocation, and GenesisCoinDistribution aren't set as there is no such thing as a "sane default" for these variables
```
GlenDC: Should just be a godoc as part of the comment above the function. So it's very visible, as it concerns the user for sure.

types/constants.go
```diff
@@ -50,200 +50,153 @@ var (
...
-		bso.UnlockHash.LoadString("b5e42056ef394f2ad9b511a61cec874d25bebe2095682dd37455cbafed4bec15c28ee7d7ed1d")
-		GenesisBlockStakeAllocation = append(GenesisBlockStakeAllocation, bso)
-		co.UnlockHash.LoadString("b5e42056ef394f2ad9b511a61cec874d25bebe2095682dd37455cbafed4bec15c28ee7d7ed1d")
-		GenesisCoinDistribution = append(GenesisCoinDistribution, co)
-	}
+// ChainConstants is a utility struct which groups together the chain configuration
+type ChainConstants struct {
+	BlockSizeLimit   uint64
+	RootDepth        Target
+	BlockFrequency   BlockHeight
```
GlenDC: Please provide properties for each property, explaining what they do

pkg/daemon/networks.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/rivine/rivine/types"
+)
+
+var (
+	// networks is the mapping of names to the parameters required to initialize them
+	networks = make(map[string]Network)
+)
+
+// Network are variables for a particular chain. Currently, these are genesis constants and bootstrap peers
+type Network struct {
```
GlenDC: I would call that NetworkConfig.

LeeSmet:
Added a WIP for feedback: 

- Replaced the exported global vars in the types package with an unexported cfg struct as some of the function in types rely on it.
- Inject the chain constants struct in all modules that require them (basically all)
- Inject the bootstrappeers in the gateway module
- Fix tests (fml)

For now I've replaced the onecoin reference in client with an injected var when the client is created since the chainconstants are now defined in the context of a network, which is defined only in the daemon. Will check tomorrow if there is a better solution for this but at least it compiles for now

GlenDC:
I'll make sure this is today around lunch-afternoon merged, as I think this is the last remaining thing of milestone `0.5`.

PR #136, opened 2018-03-07
Issue 119: Don't use seed as password

lafont-e:
We did not want to use the seed as a password initializing a
wallet when the user did no provide one.

So we have removed the password -p flag, now every wallet initialization
will ask the user for a password and a verification.

In my machine setting the password complains with **Error when encrypting wallet: error when calling /wallet/init: The number of bits must be divisible by 32** but the wallet is initialized and with the password selected.

### DIFF
--- a/pkg/client/walletcmd.go
+++ b/pkg/client/walletcmd.go
@@ -175,24 +175,32 @@ func Walletaddressescmd() {
 // Walletinitcmd encrypts the wallet with the given password
 func Walletinitcmd() {
 	var er api.WalletInitPOST
-	qs := fmt.Sprintf("dictionary=%s", "english")
-	if initPassword {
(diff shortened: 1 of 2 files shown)


pkg/client/walletcmd.go
```diff
@@ -175,24 +175,32 @@ func Walletaddressescmd() {
...
 	}
-	err := PostResp("/wallet/init", qs, &er)
+
+	repasswd, err := speakeasy.Ask("Reenter password: ")
+	if err != nil {
+		Die("Reading password failed:", err)
+	}
+
+	if repasswd != password {
+		Die("Passwords does not match !!")
```
GlenDC: `Passwords do not match`*, it's "do" in this case, not "does"

PR #139, opened 2018-03-08
add data replication module and feature

### LINKED ISSUE #120: Provide the ability to replicate specific data on an external redis-protocol KV storage
There still has to be a lot defined in this issue, but we can already outline the goal:

> Be able to filter on one or multiple addresses, and store the data (mapped to those addresses) in a KV storage. The storage is externally and accessed via [the Redis protocol](https://redis.io/topics/protocol).

By default it would do this for the entire blockchain (for which the addresses match), but I think it shouldn't be too difficult to support an optional min/max blockheight as well, should this be desired.

The original idea was to simply add it as an optional feature to the explorer module, given that it is already doing partially the work that has to be done. However @robvanmieghem came up with the idea, that we could just turn this into a separate binary, which uses the consensus and gateway modules, and which would run as a separate daemon service just for this purpose. I quite like that idea as well.

Also to be clear, for now there is no requirement to filter on data, I think the data should be considered arbitrary data, and filtering only ever happens based on addresses (sender and/or receiver(?)).

Prior to this issue #117 has to be resolved in master.

LeeSmet:
closes #120


modules/datastore/redis.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"errors"
+
+	"github.com/go-redis/redis"
+	"github.com/rivine/rivine/modules"
+	"github.com/rivine/rivine/types"
+)
+
+const (
+	stateSet         = "dbstate"
```
GlenDC: Try to use one-two letter field names, as it seems we want to keep things as small as possible.

pkg/daemon/daemon.go
```diff
@@ -74,6 +79,10 @@ func DefaultConfig() RivinedCfg {
 		Profile:    false,
 		ProfileDir: "profiles",
 		RivineDir:  "",
+
+		RedisAddr:     "localhost:6379",
```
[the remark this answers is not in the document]
LeeSmet: Since its a separate module, this module needs to be manually activavated(adding `d` to the modules string), so then a local redis on the same machine seems like a good default no?

modules/datastore/redis.go
LeeSmet: Sure. This will all change anyway with the subscriptions

pkg/daemon/daemon.go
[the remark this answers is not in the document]
GlenDC: I guess it is. For now you can leave it in.

modules/datastore/namespacemanager.go
```diff
@@ -0,0 +1,114 @@
...
+	return nsm.DB.SaveManager(nsm)
+}
+
+// Delete removes the namespace manager from the database
+func (nsm *NamespaceManager) Delete() error {
+	return nsm.DB.DeleteManager(nsm)
+}
+
+// SubscribeCs subscribes this namespace manager to the given consensus set
+func (nsm *NamespaceManager) SubscribeCs(cs modules.ConsensusSet) {
```
GlenDC: why is this a thing? I might miss something, but I fail to see a scenario where we would want to subscribe to a new consensusset? Shouldn't we just do this once (and only once) at creation time of the namespace manager?

modules/datastore/redis.go
```diff
@@ -0,0 +1,177 @@
...
+// NewRedis creates a new redis struct which attempts to connect to the specified DB/instance
+// The connection always used tcp
+func NewRedis(addr, password string, db int) (*Redis, error) {
+	client := redis.NewClient(&redis.Options{
+		Addr:     addr,
+		Password: password,
+		DB:       db,
+	})
+	// Test the connection
+	_, err := client.Ping().Result()
```
GlenDC: I think you can/should remove this, as the database interface exposes an interface func Ping, which we call ourselves already, so having it here as well, is in best case double the effort

modules/datastore/redis.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// StoreData stores data in an HSET
+func (rd *Redis) StoreData(ns Namespace, ID DataID, data []byte) error {
+	_, err := rd.cl.HSet(ns.String(), string(ID), data).Result()
+	return err
+}
+
+// DeleteData removes data from an HSET
+func (rd *Redis) DeleteData(ns Namespace, ID DataID) error {
```
GlenDC: `DeleteData` should be removed, no? I was fairly sure that we agreed upon making our replicated data immutable?

modules/datastore/redis.go
```diff
@@ -0,0 +1,177 @@
...
+					return
+				}
+				ev, ok := parsePayload(msg)
+				if !ok {
+					continue
+				}
+				// Don't waste time here
+				go fn(ev)
+			default:
+				time.Sleep(pollPeriod)
```
GlenDC: why do we do this?

modules/datastore/redis.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// DeleteData removes data from an HSET
+func (rd *Redis) DeleteData(ns Namespace, ID DataID) error {
+	_, err := rd.cl.HDel(ns.String(), string(ID)).Result()
+	return err
+}
+
+// Subscribe starts a subscription to the
+func (rd *Redis) Subscribe(fn SubEventCallback) {
```
GlenDC: you miss a `contect.Context` here, as to be able to get out of this goroutine

modules/datastore/redis.go
```diff
@@ -0,0 +1,177 @@
...
+				if msg == nil {
+					// Read a nil msg, so the channel is closed
+					return
+				}
+				ev, ok := parsePayload(msg)
+				if !ok {
+					continue
+				}
+				// Don't waste time here
+				go fn(ev)
```
GlenDC: if you do that already, than you should work with channels, working with a callback is faster, but if you execute the callback on a goroutine it's going to not save you all that much, if any, time in the end, so in that case you should go for the channel approach, which will also make the entire process thread-safe by default

pkg/daemon/commands.go
```diff
@@ -105,6 +105,13 @@ Explorer (e):
 	The explorer requires the consensus set.
 	Example:
 		%[1]sd -M gce
+DataStore (d):
+	The datastore provides an interface to connect to a (possibly remote)
+	redis instance. Data added to transactions is replicated to an HSET for every
+	unlockhash which receives coins in the transaction.
```
GlenDC: this comment is out of date, as now we do it based on the namspace ID, rather than the unlock hash

modules/datastore/database.go
```diff
@@ -0,0 +1,23 @@
+package datastore
+
+// Database is the common interface which must be implemented to be compatible with the datastore
+type Database interface {
```
GlenDC: I would remove the `Save/Get/Delete Manager` functions, as those don't belong here. That you work with managers doesn't even have to be obvious (other than the fact that you document it and expose the type, but they certainly do not belong in the func API). When you subscribe/unsubscribe you would add/delete those managers automatically. I do get why you do it, but I start to think that you might want to keep Namespace managers private to the database struct, as there is also the location where you serialize/deserialize their state AFAIK. Right now you made weird design choices to somehow make it work to get the namespacemanagers within the Datastore struct, but unless I missed something, I don't understand any benefit, while I do see it making the code get pretty messy already.

Ping and StoreData is fine. I thought we agreed to remove DeleteData?

Subscribe works a bit weird now, should also have a context, otherwise how the hell will you know the owner is done and wants the goroutine to quite. Remember rivine is supposed to be a library, so we cannot assume that the user wants to run this always the entire main goroutine. Further make a choice, either you want the callback to block, or go for a goroutine, but it's a weird design choice to use a callback but still spawn a goroutine for it each time. You also miss an Unsubscribe function by the way, unless I missed something.

modules/datastore/namespacemanager.go
```diff
@@ -0,0 +1,114 @@
...
+	// manager in the datastore, (either loading or a new subscription), so this should be fine.
+
+	err := cs.ConsensusSetSubscribe(nsm, nsm.State.RecentChangeID)
+	if err != nil {
+		nsm.log.Severe("Namespace manager failed to subscribe to consensus set: ", err)
+	}
+}
+
+// UnSubscribeCs removes this namespace manager from the previously subscribed consensus set
+func (nsm *NamespaceManager) UnSubscribeCs() {
```
GlenDC: I don't see the point of this func, should just be part of the close function, try to keep your API surface as small as possible when designing

modules/datastore/redis.go
LeeSmet: I left it here as as a precaution, if a fork happens which is larger (more blocks) than the buffer, it will need to be able to delete data. If not, the database will contain data which should have been reverted. In practice, this likely means that the data id will be faulty, and that the "actual" data id will also be present. So even though in practice we should not need to call it, I still think its usefull to have as a safety measure

doc/datareplication.md
```diff
@@ -0,0 +1,61 @@
...
+In order to allow people to store data on the blockchain, a command is exposed in the `client` which creates a minimal transaction to the provided address, while paying the miner fee, and adding the provided data. This command silently adds a specifier (as to not have the transaction rejected by the `transactionpool` for containing invalid data). The command is as follows:
+
+```bash
+rivinec wallet registerdata [destination address] [data]
+```
+
+The destination address can be any valid address (inlcuding one of your own addresses). Keeping in mind the structure the data should have, as well as the fact that the client adds the specifier by itself, we only need to add the namespace to the data. For example, if we want to write "testdata" to the namespace "1111", we can do so with the following command:
+
+```bash
+rivinec wallet registerdata 53d6162d18db2dffbe7f9dd9fcbce1c34... 1111testdata
```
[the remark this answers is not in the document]
LeeSmet: While that makes sense, I think it is more appropriate that this is moved to its own issue as the CLI is not part of this PR

modules/datastore/redis.go
GlenDC: ok, 0-db also can delete data apparently, they however don't support hash sets, so I don't think we'll ever be able to use it as we use Redis. For now let's not worry about it.

modules/datastore/database.go
LeeSmet: As to why DeleteData is here, see my reply on your earlier comment on the actual implementation.

An Unsubscribe function would imply that its possible to change the channel to which you subscribe, or subscribe to multiple, both of which are not possible. Unsubscribe is now implicit when the Datastore is closed (and thus the DB connection). 

As for save/get/delete manager functions, I suppose you mean that the datastore should handle the serialization, and just write the serialized state like any other piece of data? That would require keeping the DeleteData method btw, and possibly a GetAllDataFromNamespace (or something like that)

modules/datastore/redis.go
LeeSmet: Leftover from an earlier version which did not get cleaned up, mb

modules/datastore/database.go
```diff
@@ -0,0 +1,23 @@
...
+	// DeleteManager removes a manager in case it unsubscribes
+	DeleteManager(*NamespaceManager) error
+	// StoreData stores the specified data, linked to the namespace and the generated ID
+	// Multiple section of data must be storeable for the same namespace. The ID is unique
+	// for the namespace, but other namespaces might have the same ID's for different data
+	StoreData(Namespace, DataID, []byte) error
+	// DeleteData removes the data specified by the namespace and ID.
+	DeleteData(Namespace, DataID) error
+	// Subscribe continuously manages a channel for messages
+	Subscribe(SubEventCallback)
```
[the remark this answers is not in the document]
LeeSmet: Do we want to be able unsubscribe? I guess that makes more sense when working with channels instead of a callback

modules/datastore/redis.go
LeeSmet: The routine should end when the database is closed (as this will close the channel thus triggering the return). Since we subscribe until the database closes (for now), does it make sense to have a separate context for this?

modules/datastore/database.go
GlenDC: > An Unsubscribe function would imply that its possible to change the channel to which you subscribe, or subscribe to multiple, both of which are not possible. Unsubscribe is now implicit when the Datastore is closed (and thus the DB connection).

I don't think it implies that, unless I am confused here. But the redis user can unsubscribe a namespace, no? At least they should be able to do that.

modules/datastore/database.go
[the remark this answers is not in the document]
GlenDC: Has nothing to do with the channel he, even when you kept using a callback, that would be desired.

modules/datastore/redis.go
GlenDC: Perhaps not. You are right. I also see now that I am confused. You do not need an Unsubscribe feature here indeed for this. Subscribing is enough. But in that case, I would call it `SetEventCallback`, or something like that.

modules/datastore/database.go
LeeSmet: ah i see, you are thinking about subscribing to a namespace. This function deals with the subscription to the redis replication channel. Once a message is picked up from the replication channel, it gets parsed, and then a new routine is spawned in which a new manager is created and subscribes to the cs (after which that routine dies), or an existing manager is deleted (after which the routine also dies)

modules/datastore/database.go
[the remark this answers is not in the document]
LeeSmet: I suppose you are thinking that this subscribe means the subscription of a new namespace, while this subscribes the database to the replication channel to listen for namespace subscriptions as pointed out in your/my other comment

modules/datastore/database.go
GlenDC: yeah I was thinkiing that indeed. However I still think that:

+ the handling the manager should happen behind the scenes of the database, no need to expose that AFAIK;
+ when creating the manager we can just give the consensus set already. So that in the constructor we just subscribe automatically and during Close we can unsubscribe, I don't see the point of exposing that either;

modules/datastore/database.go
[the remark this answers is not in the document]
GlenDC: yup, so nvm

modules/datastore/database.go
LeeSmet: I presume you mean handling the managers should happen in the datastore

GlenDC: I don't think so, as it makes the entire API a lot more complex. in the end the database anyhow handles the subscribe/unsubscribe stuff using those event handlers. The database also supposed to be serializing/deserializing the state, so in the end, I don't get the point of having it in the datastorage struct at all. In fact, that even means that this datastorage struct doesn't even has to know that NameSpaceManagers are used at all.

LeeSmet: The database(redis) only parses the messages from the subscribe channel, and if a valid message is found, it invokes the callback. So the actual managment (creating and removing) of managers is handled by the Datastore(module). Its true that the serialization is currently handled by the database, but as you pointed out that could be handled by the datastore as well, and then use a savedata/deletedata command which is already provided

doc/datareplication.md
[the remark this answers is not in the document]
GlenDC: issue created #175

GlenDC:
Looks good in general I think.

+ going to rebase master;
+ going to play with a bit;
+ going to check if I can/need-to improve some stuff while playing with it;
+ will add issues for stuff that is still left to be done;

PR #144, opened 2018-03-13
update go-bip39 lib, integrate it and improve code around it

GlenDC:
don't squash


GlenDC:
@robvanmieghem unit tests pass now. Was using assert by accident without vendoring it. And it seemed silly to do so for just one test, so I removed the need for it, by using plain old std `Testing`.

GlenDC:
Yeah I agree with many of those things @robvanmieghem. Perhaps it's stuff that can better wait until we have #155 resolved or wontfix-closed though, as that's another big impact feature in this area. After that we can and should indeed spend some time cleaning up. Supporting unseeded, I have no real opinion about that yet, as I haven't read on that subject, but it sounds a feature that might come in handy indeed.

Could you perhaps make one or multiple issues to track and backlog the further improvements you have in mind (optionally linking those with #155). As I think we can make the wallet code quite comprehensive, clean and proper with a bit of more work in regards with such issues. But just mentioning it in a comment, is going to be forgotten eventually.

robvanmieghem:
https://github.com/rivine/rivine/issues/158
https://github.com/rivine/rivine/issues/159

PR #157, opened 2018-03-15
Fix the progress estimation when syncing the blockchain

lafont-e:
We have changed the start of the net to a time where the number of blocks is known

The code was set to start in 2016, so the estimation was wrong as it calculated like 100k records

### DIFF
--- a/pkg/client/consensuscmd.go
+++ b/pkg/client/consensuscmd.go
@@ -41,7 +41,7 @@ Target: %v
 		}
 		fmt.Printf(`Synced: %v
 Height: %v
-Progress (estimated): %.f%%
+Progress (estimated): %.02f
 `, YesNo(cg.Synced), cg.Height, estimatedProgress)
 	}
 }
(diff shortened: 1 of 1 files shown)


pkg/client/consensuscmd.go
```diff
@@ -50,8 +50,12 @@ Progress (estimated): %.f%%
...
 // the past and dividing by 10 minutes (the block time).
 func EstimatedHeightAt(t time.Time) types.BlockHeight {
-	block5e4Timestamp := time.Date(2016, time.May, 11, 19, 33, 0, 0, time.UTC)
-	diff := t.Sub(block5e4Timestamp)
-	estimatedHeight := 5e4 + (diff.Minutes() / 10)
-	return types.BlockHeight(estimatedHeight + 0.5) // round to the nearest block
+	// This should the timestamp of the genesis block, Not sure why it is hardcoded
+	// but as the tfchain has been stoped, the time generates much more records than
+	// the ones available, so we start from a know place
+	startTimestamp := time.Date(2018, time.March, 15, 13, 0, 0, 0, time.UTC)
```
GlenDC: Replace this with `startTimestamp := types.GenesisTimestamp`

pkg/client/consensuscmd.go
```diff
@@ -50,8 +50,12 @@ Progress (estimated): %.f%%
...
+	diff := t.Sub(startTimestamp)
+	// Estimated number of blocks, starting from 15/03/2018 there were 2660
+	estimatedHeight := 2600 + (diff.Minutes() / 10)
```
GlenDC: We cannot hardcode it like this (the block count), it has to be somehow known. The best would be is if we provide a `explorer` endpoint to know the height, such that we can expose a public GET REST Endpoint on the tfchain dns. For example for bitcoin you can do https://blockchain.info/q/getblockcount (perhaps you also have something more official) but this might be a solution. As I don't think it is possible to do it otherwise. Unless we can ask the blockchain ourselves from our peers.

pkg/client/consensuscmd.go
lafont-e: I do, but you know the estimation would be wrong the Genesis timestamp starts on February 2015

pkg/client/consensuscmd.go
lafont-e: I agree, this is not cool, we need a function that returns the number of blocks in the chain, rivine have the Height() method but only return the height of the consensus, not the total number of blocks in the chain when synced.
I've researched a little and the only solution I can imagine is to count the blocks arriving, but it's ugly as hell, the correct one, although intrusive, is to create a Stats() method in consensusset (both interface and struct) that return this kind of info and another one we may need or like, also we must add an api call to return such info

pkg/client/consensuscmd.go
GlenDC: that genesis timestamp is set by the blockchain creator. So in tfchain we'll be able to set this property later this week. Thus all is good there.

GlenDC:
Going to close this PR as the things that have to be done to solve the issue is trying to solve is blocked by missing features that we have no time for now.

PR #160, opened 2018-03-15
add transaction short ID and consensus-get-transaction func

### LINKED ISSUE #149: implement and support short+unique+ordered Transaction IDs
Each transaction should be identifiable with a unique ID, which is both short enough (8 bytes) as well as unique (meaning only ever transaction at a given  time should have that ID) and is ordered (meaning that the higher the ID the later in the blockchain (and block) the transaction should be. In that sense the ID is linear and increasing with time.

The format I propose to go for (based on @LeeSmet's original proposal) is:

```
[ block height (50 bits) | local-to-block transaction ID (14 bits) ]

(!) (!) (!)
It's important that these integers get serialized using binary-little-endian,
and in the order as proposed, just that the resulting byte slice from left to right
gives the expected ordered sequence for all transactions.
(!) (!) (!)
```

This would mean we support up to `16384` transactions per block (which is more than we ever need). According to some charts for  bitcoins there aren't really more than `2710` transactions per block in their chain with the average being closer to `1600`. Taken into account that the amount of transactions per block grow with the growth of the network, we would still have a lot of growth with a max of `4096` transactions per block. And given that bitcoin is at this point probably way more popular than we'll ever get, it seems that with `14` bits we're fine, even we decide to want to allow more transactions for whatever reason.

Now, it is true that Rivine's BlockHeight type allows for `64` bits, while here we only allow `50` bits. However, one should realize that only `52` bits does still allow for up to `1.125.899.906.842.624` blocks PER blockchain. Given that the most popular blockchain, main Bitcoin, is not only popular but also has been running for many years, and still only has `513483` at the time of writing (can be checked using https://blockchain.info/q/getblockcount AFAIK), I would be really surprised that this would ever become a problem.

With these numbers in mind, it would mean we can handle a total amount of transactions between `56.294.995.342.131.200` and `1688849860263936000`, given an average transaction/block count between `50` and `1500`. Given such numbers, I would say we're pretty safe with a 64 bit ID.

It is also important that `rivined` exposed in the Golang and REST API a function/endpoint to get a transaction based on this short ID. Could be on endpoint `GET /transactionpool/transaction/:shortid`, but open for other endpoint suggestions as well, should a different namespace be desired.

GlenDC:
+ add TransactionShortID type (closes #149);
+ add GetTransactionAtShortID function to ConsensusSet:
  + this is also exposed as the '/consensus/transactions/:shortid' endpoint;
  + the newly exposed endpoint is also available as
    'rivinec consensus transaction <shortID>';

### DIFF
--- a/pkg/client/consensuscmd.go
+++ b/pkg/client/consensuscmd.go
@@ -1,7 +1,10 @@
 package client
 
 import (
+	"encoding/json"
 	"fmt"
+	"os"
+	"regexp"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -17,9 +20,16 @@ var (
 		Long:  "Print the current state of consensus such as current block, block height, and target.",
 		Run:   wrap(Consensuscmd),
 	}
+
+	consensusTransactionCmd = &cobra.Command{
+		Use:   "transaction <shortID>",
+		Short: "Get an existing transaction",
+		Long:  "Get an existing transaction from the blockchain, using its given shortID.",
+		Run:   wrap(Consensustransactioncmd),
+	}
 )
 
-// Consensuscmd is the handler for the command `siac consensus`.
+// Consensuscmd is the handler for the command `rivinec consensus`.
 // Prints the current state of consensus.
 func Consensuscmd() {
 	var cg api.ConsensusGET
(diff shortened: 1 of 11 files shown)


pkg/client/consensuscmd.go
```diff
@@ -55,3 +65,28 @@ func EstimatedHeightAt(t time.Time) types.BlockHeight {
...
+	}
+
+	var txn api.ConsensusGetTransaction
+	err := GetAPI("/consensus/transactions/"+shortID, &txn)
+	if err != nil {
+		Die("failed to get transaction: ", err, "; shortID: ", shortID)
+	}
+
+	encoder := json.NewEncoder(os.Stdout)
+	//encoder.SetIndent("", "  \t")
```
LeeSmet: Does this comment serve any purpose or can it be deleted?

GlenDC: No it can be removed, I was first thinking to make it indented, but than I changed my mind as it makes piping harder, could make it optional with a flag, but so far we don't use any flags AFAIK, so commented it. Will just remove it for now.

PR #207, opened 2018-03-28
merge dev 0.6 (transactions 2.0) into master

GlenDC:
don't squash


api/wallet.go
```diff
@@ -57,10 +58,25 @@ type (
 		PrimarySeed string `json:"primaryseed"`
 	}
 
+	// WalletTransactionPOST contains the unlockhash and amount of money to send,
+	// during a POST call to /wallet/transaction, funding the output,
+	// using available inputs in the wallet.
+	WalletTransactionPOST struct {
+		UnlockHash types.UnlockHash `json:"unlockhash"`
+		Amount     string           `json:"amount"`
```
LeeSmet: wouldn't it be possible to change the type of `Amount` to `types.Currency`? Since that type already implements the JSON marshaller interface.

api/wallet.go
```diff
@@ -293,6 +316,53 @@ func (api *API) walletSeedsHandler(w http.ResponseWriter, req *http.Request, _ h
...
+}
+
+// walletTransactionCreateHandler handles API calls to POST /wallet/transaction.
+func (api *API) walletTransactionCreateHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var body WalletTransactionPOST
+	if err := json.NewDecoder(req.Body).Decode(&body); err != nil {
+		WriteError(w, Error{"error decoding the supplied transaction output: " + err.Error()}, http.StatusBadRequest)
+		return
+	}
+	amount, ok := scanAmount(body.Amount)
```
LeeSmet: related to previous comment, this check could maybe be avoided if we decode directly into a `types.Currency` ?

modules/explorer.go
```diff
@@ -23,15 +23,13 @@ type (
...
-		ArbitraryDataCount        uint64 `json:"arbitrarydatacount"`
-		TransactionSignatureCount uint64 `json:"transactionsignaturecount"`
+		MinerPayoutCount      uint64 `json:"minerpayoutcount"`
+		TransactionCount      uint64 `json:"transactioncount"`
+		CoinInputCount        uint64 `json:"coininputcount"`
+		CoinOutputCount       uint64 `json:"coinoutputcount"`
+		BlockStakeInputCount  uint64 `json:"blockstakeinputcount"`
+		BlockStakeOutputCount uint64 `json:"blockstakeoutputcount"`
+		MinerFeeCount         uint64 `json:"minerfeecount"`
 	}
```
LeeSmet: Is there any particular reason why we no longer care for the amount of arb data in a block? Or is this info just deemed not usefull/confusing/...

pkg/client/walletcmd.go
```diff
@@ -247,46 +245,45 @@ func Walletseedscmd() {
...
+// walletregisterdatacmd registers data on the blockchain by making a minimal transaction to the designated address
 // and includes the data in the transaction
-func Walletregisterdatacmd(namespace, data, dest string) {
-	// At the moment, we need to prepend the non sia prefix for the transaction to be accepted by the transactionpool
-	dataBytes := append(modules.PrefixNonSia[:], []byte(namespace)...)
-	dataBytes = append(dataBytes, []byte(data)...)
-	encodedData := base64.StdEncoding.EncodeToString(dataBytes)
-	err := Post("/wallet/data", fmt.Sprintf("destination=%s&data=%s", dest, encodedData))
+func walletregisterdatacmd(namespace, dest, data string) {
+	// / At the moment, we need to prepend the non sia prefix for the transaction to be accepted by the transactionpool
```
LeeSmet: seems like this comment can go as well

types/unlockhash.go
```diff
@@ -38,7 +108,7 @@ func (uh UnlockHash) MarshalJSON() ([]byte, error) {
 // that has been encoded to a hex string.
 func (uh *UnlockHash) UnmarshalJSON(b []byte) error {
 	// Check the length of b.
-	if len(b) != crypto.HashSize*2+UnlockHashChecksumSize*2+2 && len(b) != crypto.HashSize*2+2 {
+	if len(b) != (crypto.HashSize+UnlockHashChecksumSize)*2+4 {
```
LeeSmet: This is just my opinion (putting in my 2 cents), but I feel this check would be more readable as (1+crypto.HashSize + UnlockHashChecksumSize) * 2 + 2. For me it seems like that would more obviously convey that the length should be equal to the length of the UnlockHash in byteform * 2 for the encoding, + 2 for the `"` characters.

Since now it took me some time to realize why the `+4` was not just `+2`, but everyone probably has a different opnion about this

types/unlockhash.go
```diff
@@ -47,29 +117,38 @@ func (uh *UnlockHash) UnmarshalJSON(b []byte) error {
...
-// LoadString loads a hex representation (including checksum) of an unlock hash
-// into an unlock hash object. An error is returned if the string is invalid or
+// LoadString loads a hex representation (including checksum)
+// of an unlock hash into an unlock hash object.
+// An error is returned if the string is invalid or
 // fails the checksum.
 func (uh *UnlockHash) LoadString(strUH string) error {
 	// Check the length of strUH.
-	if len(strUH) != crypto.HashSize*2+UnlockHashChecksumSize*2 {
+	if len(strUH) != 2+crypto.HashSize*2+UnlockHashChecksumSize*2 {
```
LeeSmet: Maybe for consistency the calculation of the expected length could have the same format as the one above?

api/wallet.go
GlenDC: *unmarshaller though, but yeah you're right

modules/explorer.go
GlenDC: It wasn't counting the amount of arb data in a block, it was amount of arb data _elements_ in a single transaction. Note that it was counting elements not individual bytes. Since this PR the arb data is however just a 1D array and thus always a single element. Hence I don't see a reason for it anymore, given that it will always be either zero or 1. We could make it into a boolean property I suppose. We could also count the amount of encoded arbitrary data in byte count. All possible.

types/unlockhash.go
GlenDC: You're right, that formula does take a bit of time to sink in before one might realize why it is like that. Will reshape it a bit, as you suggest. And I'll also add some comments by it to explain where the numbers come from, just to be sure.

modules/explorer.go
GlenDC: Ah no. Looking at it again, it seems to be about block facts, and thus it could still make sense I suppose. Will add it once again. Not sure if it's really useful though.

PR #271, opened 2018-04-12
Fork recovery

LeeSmet:


modules/consensus/consensusset.go
```diff
@@ -236,6 +236,31 @@ func (cs *ConsensusSet) BlockHeightOfBlock(block types.Block) (height types.Bloc
 	return height, exists
 }
 
+// FindParentBlock finds the parent of a block at the given depth. While this function is expensive, it guarantees
+// that the correct parent block is found, even if the block is not on the longest fork. If it is known up front
```
GlenDC: How come a parent can be the correct parent block, if not the longest chain?

modules/consensus/accept.go
```diff
@@ -151,6 +151,9 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
 		}
 		var revertedBlocks, appliedBlocks []*processedBlock
 		revertedBlocks, appliedBlocks, err = cs.forkBlockchain(tx, newNode)
+		if len(revertedBlocks) != 0 {
```
GlenDC: we don't want to log if applied block isn't `0`?

modules/consensus/consistency.go
```diff
@@ -117,7 +117,7 @@ func (cs *ConsensusSet) checkRevertApply(tx *bolt.Tx) {
 		manageErr(tx, err)
 	}
 	if consensusChecksum(tx) != parent.ConsensusChecksum {
-		//manageErr(tx, errors.New("consensus checksum mismatch after reverting")) //this fails issue #28
+		manageErr(tx, errors.New("consensus checksum mismatch after reverting"))
```
GlenDC: if this doesn't fail any longer, I guess that's a good sign

modules/consensus/consensusset.go
LeeSmet: being on the longest chain/active fork has nothing to do with being a correct/valid parent block. If we have a block `b`, then said block has a parent block `pb`. If we know the block height (`hb`) of `b`, then the "i-th generation parent" of `b` is at blockheight `hb - i`. But the consensus set db only index block by blockheight if they are on the currently active fork / longest chain. So if `b` is currently not on the active chain, then it's possible that it's parent is also not on the longest chain, and therefore not indexed. So the only way to get the correct block is then to "manually" traverse the chain by getting each block, looking at the `parentID` in the header, get that block, look at the `parentID` in its header, etc... untill we went back `i` blocks. Now we have the "i-th generation parent" of our block, even if its not on the currently active fork/longest chain.

PR #303, opened 2018-04-30
update consensus/explorer db versions

### LINKED ISSUE #301: ensure we can unmarshal all legacy outputs/inputs
We persist inputs and outputs not just as part of a Transaction, but also in other types of structures (e.g. diffs). This makes it that the current master code is broken for existing chains, as they wouldn't be able to load those diffs and thus result in panics.

An example of such a struct is https://github.com/rivine/rivine/blob/9efb029aefd6f257c4c9eac1cf28f64c163f5806/modules/consensus/processedblock.go#L24-L40 (which contains diffs, which contains outputs).

It is important that those places where we already serialized data, have the logic to deserialize any values previously serialized with full legacy compatibility.

GlenDC:
closes #301

ensure that all buckets have the data in the new format

### DIFF
--- a/modules/consensus/persist_legacy.go
+++ b/modules/consensus/persist_legacy.go
@@ -1,0 +1,253 @@
+package consensus
+
+import (
+	bolt "github.com/rivine/bbolt"
+	"github.com/rivine/rivine/build"
+	"github.com/rivine/rivine/crypto"
+	"github.com/rivine/rivine/encoding"
+	"github.com/rivine/rivine/modules"
+	"github.com/rivine/rivine/persist"
+	"github.com/rivine/rivine/types"
+)
+
+// convertLegacyDatabase converts a 0.5.0 consensus database,
+// to a database of the current version as defined by dbMetadata.
+// It keeps the database open and returns it for further usage.
+func convertLegacyDatabase(filePath string) (db *persist.BoltDatabase, err error) {
+	var legacyDBMetadata = persist.Metadata{
+		Header:  "Consensus Set Database",
+		Version: "0.5.0",
+	}
+	db, err = persist.OpenDatabase(legacyDBMetadata, filePath)
+	if err != nil {
+		return nil, err
+	}
+	err = db.Update(func(tx *bolt.Tx) error {
+		if bucket := tx.Bucket(BlockMap); bucket != nil {
+			if err := updateLegacyBlockMapBucket(bucket); err != nil {
+				return err
+			}
+		}
+		if bucket := tx.Bucket(CoinOutputs); bucket != nil {
+			if err := updateLegacyCoinOutputBucket(bucket); err != nil {
+				return err
+			}
+		}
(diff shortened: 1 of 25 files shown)


modules/consensus/persist_legacy.go
```diff
@@ -0,0 +1,250 @@
...
+	bolt "github.com/rivine/bbolt"
+	"github.com/rivine/rivine/build"
+	"github.com/rivine/rivine/crypto"
+	"github.com/rivine/rivine/encoding"
+	"github.com/rivine/rivine/modules"
+	"github.com/rivine/rivine/persist"
+	"github.com/rivine/rivine/types"
+)
+
+func openLegacyDatabase(filePath string) (db *persist.BoltDatabase, err error) {
```
robvanmieghem: I would give  name this function a name that includes "convert" or "update" since  this is what it does.

modules/consensus/persist_legacy_test.go
```diff
@@ -0,0 +1,67 @@
...
...00000000000000000000000000000000150d00000000000000000004cba26d4d353bcf0219f4f0257f817e90d8864f16a4058fd6bd5b701b00002d9bcf55acdf7eb91dbd632fa5f441e2fe68a592704a3c09419b8294b93201010000000000000001b530693498b9968cf78c0a3e796d573c64c6989a05f698964c5d9d6b2e09da45050000000000000002540be40001e34588bee49b2cbd53f2198cd5022fbbe78aecb8125a39efb8699720b946e84e020000000000000000441deb06592694fb0f7fb41dd9ec7239b01cf83ec0100f92c5b3312ebf5e9fec020000000000000003e80188adad4e890b57afdcea6ea7c7d8dbc42c939801a3a8015cfc4e779d57d34cf4019caaf5e1ed6aa1bce252aba995d7e14c82005f3da15bf08a8425d9c86eedafcf020000000000000003e80188adad4e890b57afdcea6ea7c7d8dbc42c939801a3a8015cfc4e779d57d34cf4020000000000000001707a8d2f97182646caeb2404e5fdd59119a3d2de4d2f40efbde4e3dca8f7d132050000000000000002540be4000188adad4e890b57afdcea6ea7c7d8dbc42c939801a3a8015cfc4e779d57d34cf4e50f00000000000000b530693498b9968cf78c0a3e796d573c64c6989a05f698964c5d9d6b2e09da45050000000000000002540be40001e34588bee49b2cbd53f2198cd5022fbbe78aecb8125a39efb8699720b946e84e150d000000000000010000000000000001ec6620d8c7813cc86ae5c6021d4bb27f0fd77b4f8a4d98f2844032e64d071f4f0040450300000000beeb48fb26bd7cd43e97d0a1a974241b5c071bea3c736fec07d6685da6e39213`,
```
robvanmieghem: Would be nice to have the data in a human readable format (json even)  as a comment so one can read the actual testcase.

PR #325, opened 2018-05-18
Multisig support

LeeSmet:
Support for multisig output and inputs
Ability to provide a json format txn to the client to be committed to the transactionpool (`wallet send transaction <txnjson>`)


types/unlockcondition_test.go
```diff
@@ -578,6 +614,109 @@ func TestUnlockConditionEqual(t *testing.T) {
...
+				},
+			},
+			&MultiSignatureCondition{
+				MinimumSignatureCount: 2,
+				UnlockHashes: UnlockHashSlice{
+					unlockHashFromHex("01e89843e4b8231a01ba18b254d530110364432aafab8206bea72e5a20eaa55f70b1ccc65e2105"),
+					unlockHashFromHex("01a6a6c5584b2bfbd08738996cd7930831f958b9a5ed1595525236e861c1a0dc353bdcf54be7d8"),
+					unlockHashFromHex("01a6a6c5584b2bfbd08738996cd7930831f958b9a5ed1595525236e861c1a0dc353bdcf54be7d8")},
+			},
+			"",
```
GlenDC: I think you miss a test case which tests what if minSigCount is equal, length of unlock hashes is equal, but different unlock hash(es) are used.

types/unlockcondition_test.go
```diff
@@ -2193,6 +2814,33 @@ func TestUnlockConditionFulfillable(t *testing.T) {
...
+		{
+			&MultiSignatureCondition{
+				MinimumSignatureCount: 3,
+				UnlockHashes: UnlockHashSlice{
+					unlockHashFromHex("01e89843e4b8231a01ba18b254d530110364432aafab8206bea72e5a20eaa55f70b1ccc65e2105"),
+					unlockHashFromHex("01a6a6c5584b2bfbd08738996cd7930831f958b9a5ed1595525236e861c1a0dc353bdcf54be7d8"),
+				},
+			},
+			FulfillableContext{},
+			false,
```
GlenDC: Fulfillable should always be `True`. As commented earlier (perhaps not properly documented, or badly named), but fulfillable only returns `false`, if it cannot be fulfilled because of the info given in the `FulfillableContext`. It has nothing to do with not being fulfillable because of other reasons (e.g. not being a standard condition, as is here the case).

GlenDC:
Your unit tests fail by the way @LeeSmet.

GlenDC:
@LeeSmet could you also rebase master, so that I can update the transactions documentation in `/doc/transactions/...` before I merge this PR into master (once we agree on all your commits.

PR #329, opened 2018-05-22
limit arbitrary data size to 83 bytes by default

### LINKED ISSUE #328: limit arbitrary data size to 83 bytes by default
As inspired by `OP_RETURN` from bitcoin. It should however be configurable as some chains might have a different view on this for some reason.

GlenDC:
fixes #328


LeeSmet:
Seems good, though we will need to verify that there is no txn on the chain which violates this before we release. For now, this is still fine

Issue #338, opened 2018-05-22
add create cointransaction command


GlenDC:
The new `rivinec create transaction` command will create a transaction, spending the defined inputs, spreading the inputs over the also defined outputs:

```
rivinec create cointransaction <parentID>... <dest>|<rawCondition> <amount> [<dest>|<rawCondition> <amount>]...
```

The created transaction will be returned as a JSON-encoded transaction.

+ should it not be clear from the command signature above: the command takes one or multiple parentIDs (fulfillment is selected automatically based on the condition of the parentID which is the ID of the unspent output which is to spent), as well as one or multiple condition+amount pairs, meaning you'll need some trial-and-error logic to allow for this dual-argv feature
+ an error should be returned if `sum(inputs) != sum(outputs)+transactionFee`
+ at least one input and one output has to be given
+ when creating the transaction, the input's fulfillment type is based on the condition linked to the output, identified by the given parentID;

The main motivation is to have it be another building block of our high-level multisig CLI API.

Examples of calls to this command:

```bash
# example spending one input, over two outputs
$ rivinec create cointransaction 97495f5c40d392046bd45c27acc860c6a93581930a735e0990a1e42a05cbe55e 011d1f216ebb0ae0742d7910d437d348c20f670cdb9e88ca1f88646877ace9d0255b394ed1d76d 69999.5 01907fef3ba1c3905021ae2d1486adf9bc8721821229a8a858f567b7303a26dfba454db47fa71d 30000.2

# example spending two inputs, over two outputs
$ rivinec create cointransaction 97495f5c40d392046bd45c27acc860c6a93581930a735e0990a1e42a05cbe55e 69384df8616d671c0106af5900d2240c3c20ba365def48f3afe4b9e824862e3d 011d1f216ebb0ae0742d7910d437d348c20f670cdb9e88ca1f88646877ace9d0255b394ed1d76d 69999.5 01907fef3ba1c3905021ae2d1486adf9bc8721821229a8a858f567b7303a26dfba454db47fa71d 30000.2

# example spending two inputs, over one output
$ rivinec create cointransaction 97495f5c40d392046bd45c27acc860c6a93581930a735e0990a1e42a05cbe55e 69384df8616d671c0106af5900d2240c3c20ba365def48f3afe4b9e824862e3d 01907fef3ba1c3905021ae2d1486adf9bc8721821229a8a858f567b7303a26dfba454db47fa71d 99999.7
```

[...]

Issue #337, opened 2018-05-22
add `rivinec merge inputs` command


GlenDC:
Signature:

```
rivinec merge inputs <rawTransaction1> <rawTransaction2> [rawTransactionN...] [flags]
```

Note that it will only merge where possible, and return an otherwise. Meaning that it cannot merge anything, and neither should it try to, which would affect signature output.

Knowing this, it can only merge transactions which hold all following properties:

+ define the same coin/blockstake inputs/outputs, and in the same order;
+ define the same arbitrary data, miner fees and transaction version;

While the outputs have to be exactly the same, the inputs are allowed to differ in content, as long as they define the same parentID and that the fulfillment doesn't define anything different which would change how the signature would be computed.

Knowing that, the only possible merge we would currently support is merging the pairs of MultiSignatureFulfillments. Anything else should be the same.

Issue #336, opened 2018-05-22
add `wallet sign` command


GlenDC:
The `wallet sign <rawJsonTransaction>` will sign all inputs it can sign for a given JSON-encoded raw transaction and output that same transaction JSON-encoded, signed where possible.

PR #339, opened 2018-05-22
add document which explains multisig

GlenDC:
This document explains what multisig is, links to the JSON-encoding of
the related condition/fulfillment, explains the use cases,
and also gives a few example flows one could implement to support
this feature in its rivine-enabled application.

The bash representation of the flow examples described in this document relies on the features to be implemented as part of following issues:

+ #338: create coin transactions as a raw JSON-encoded string;
+ #336: sign a raw JSON-encoded transaction and output it back in the JSON-encoded format;

As an extra it also relies on #337, as to be able to sign transactions by 3 or more parties in parallel.

Implicitly it also relies on #332, as that command will/can be used to manually, via the CLI client, find out the ID of the unspent output that is to be spend using multiple (required) signatures.

This document serves 3 purposes:

+ As documentation for anyone willing to support this feature into their rivine-enabled application (as mentioned above);
+ It motivates all feature-requests that were created today, as part of the multisig feature of the 1.0.6 milestone;
+ It ensures that we have entire flows ready in our head, such that we make sure that you spend unspent outputs locked by multisig conditions;


GlenDC:
Our repo seems to fail to build/test on Golang TIP, we might need to update some vendoring.

Issue #341, opened 2018-05-22
Show balance of multisig wallets one is part of


GlenDC:
Right now the `rivinec wallet` command doesn't show yet the balance targeted towards multisig addresses. It should.

Proposed output format:

```
Wallet status:
%s, Unlocked
Confirmed Balance:   %v
Locked Balance:      %v
Unconfirmed Delta:   %v
BlockStakes:         %v BS
[Locked BlockStakes:  %v BS] # hidden if 0

Multisig Wallets

03sdafafsdfsfsfsfdsfesresfesfsefsddadadadaddasdafsefefsfdssdsdadadadadadaddsad
Confirmed Balance:   %v
[Locked Balance:      %v] # hidden if 0
[Unconfirmed Delta:   %v] # hidden if 0
[BlockStakes:         %v BS] # hidden if 0
[Locked BlockStakes:  %v BS] # hidden if 0

033432423432424324242432424234234234234234234234234243242342342342342342434232
Confirmed Balance:   %v
[Locked Balance:      %v] # hidden if 0
[Unconfirmed Delta:   %v] # hidden if 0
[BlockStakes:         %v BS] # hidden if 0
[Locked BlockStakes:  %v BS] # hidden if 0
```

Notes:
+ Everything above the Multisig wallet output section remains as how it was;
+ The multisig wallet output section is only printed if the user's wallet actually is part of a multisig wallet, meaning there is some confirmed/locked/unconfirmed coins/blockstakes linked to an address of the user's wallet;

Using the "create multisigaddress" feature, to be implemented in #340, one could figure out easily who co-owns a certain multisig wallet shown in the output above, should that not be clear from the balance.

GlenDC:
This issue does imply some work in the wallet module to ensure that we in-memory-bucket the shared wallet balance somehow.

PR #352, opened 2018-05-25
Create and sign transaction, create multisig address 

LeeSmet:


modules/consensus/outputs.go
```diff
@@ -0,0 +1,30 @@
+package consensus
+
+import (
+	bolt "github.com/rivine/bbolt"
+	"github.com/rivine/rivine/types"
+)
+
+// GetCoinOutput returns the coin output for the given ID
```
GlenDC: returns the **unspent** coin output

modules/consensus/outputs.go
```diff
@@ -0,0 +1,30 @@
...
+		co, err = getCoinOutput(tx, id)
+		return nil
+	})
+	if dbErr != nil {
+		panic(dbErr)
+	}
+	return co, err
+}
+
+// GetBlockStakeOutput returns the blockstake output for the given ID
```
GlenDC: returns the **unspent** blockstake output

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,136 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
 	return txnSet, nil
 }
 
+// AttemptSigning tries to sign any input for which keys are loaded in the wallet
+func (tb *transactionBuilder) AttemptSigning() error {
```
GlenDC: Why is this called `AttemptSigning`, and not just `Sign`? In the end we always try no, why is it special this time?

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,136 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
+	if tb.signed {
+		return errBuilderAlreadySigned
+	}
+
+	tb.wallet.mu.Lock()
+	defer tb.wallet.mu.Unlock()
+
+	// Use an inner function so we can drill down into wrapper conditions
+	var f func(int, *types.CoinInput, types.MarshalableUnlockCondition) error
+	var g func(int, *types.BlockStakeInput, types.MarshalableUnlockCondition) error
```
GlenDC: why make inline functions of it, rather than private functions of the `*transactionBuilder`?

modules/wallet/wallet_test.go
```diff
@@ -586,3 +586,29 @@ func (css *consensusSetStub) ConsensusSetSubscribe(subscriber modules.ConsensusS
 func (css *consensusSetStub) Unsubscribe(subscriber modules.ConsensusSetSubscriber) {
 	delete(css.subscribers, subscriber)
 }
+
+func (css *consensusSetStub) GetCoinOutput(id types.CoinOutputID) (co types.CoinOutput, err error) {
+	for _, block := range css.blocks {
```
GlenDC: Isn't there a faster way? Given that the consensus contains _all_ blocks with unspent blocks, this would make it quite a slow operation, no?

modules/wallet/wallet_test.go
```diff
@@ -586,3 +586,29 @@ func (css *consensusSetStub) ConsensusSetSubscribe(subscriber modules.ConsensusS
...
+					return co, nil
+				}
+			}
+		}
+	}
+	return types.CoinOutput{}, errors.New("Coin output not found in database")
+}
+
+func (css *consensusSetStub) GetBlockStakeOutput(id types.BlockStakeOutputID) (bso types.BlockStakeOutput, err error) {
+	for _, block := range css.blocks {
```
GlenDC: Same comment as for the `GetCoinOutput` method implementation of above.

pkg/client/defaultclient.go
```diff
@@ -185,6 +187,10 @@ func DefaultCLIClient(cfg Config) {
 		walletListUnlockedCmd,
 		walletListLockedCmd)
 
+	walletCreateCmd.AddCommand(
+		walletCreateMultisisgAddress,
+		walletCreateCoinTxnCmd)
```
GlenDC: Once we agree on this PR, you could also already copy the code and make it work for the `walletCreateBlockstakeTxnCmd`, seems like small effort compared to the rest of the work done for this PR.

pkg/client/walletcmd.go
```diff
@@ -705,5 +743,86 @@ func walletlistlocked() {
 			fmt.Println()
 		}
 	}
+}
+
+func walletcreatemultisigaddress(cmd *cobra.Command, args []string) {
+	if len(args) < 3 {
```
GlenDC: no need to do this here, you can have that requirement right in the definition of your (cobra) command. See https://godoc.org/github.com/spf13/cobra#MinimumNArgs

GlenDC:
Would be awesome should you be able to wrap up this PR today, as well as start+finish issue https://github.com/rivine/rivine/issues/341. If we can get these finished today, I could play a bit with multisig tonight and this weekend at home, using our high level CLI functionality.

GlenDC:
Reported the failed CI test as issue #353, do not worry about it @LeeSmet.

modules/wallet/transactionbuilder.go
LeeSmet: The transactionbuilder already has a method `sign`. This method only works on inputs which have been added with `Fundxxx`, since those inputs are local to the wallet. This also means that no special handling is needed there as they only ever sign the singlesig fulfillments. So I thought it would be best to keep this as a separate method. And in this method, we attempt to find a suitable key to sign every input

modules/wallet/transactionbuilder.go
LeeSmet: Started out as an anonymous function, which didn't work in the end due to the recursion. And I didn't change it at the time as its only used by the enclosing method

modules/wallet/transactionbuilder.go
GlenDC: Perhaps call it than `SignAllPossibleInputs`?

modules/wallet/transactionbuilder.go
GlenDC: Well I would prefer to turn those into private methods, makes the core function much smaller and readable

modules/wallet/wallet_test.go
LeeSmet: As this is implemented on the consensussetstub, the list of blocks and transactions will never be too big imo. Any decent size test should use a real cs. If there is ever a time where this stub is used for thousands of blocks, it will need to be refactored entirely imo

modules/wallet/wallet_test.go
LeeSmet: same as above

modules/wallet/wallet_test.go
GlenDC: Yep, you're correct, sorry, didn't see that.

api/wallet.go
```diff
@@ -630,9 +632,22 @@ func (api *API) walletCreateCoinTransactionHandler(w http.ResponseWriter, req *h
...
-		Transaction: tx,
-	})
+	WriteJSON(w, tx)
+}
+
+// walletCreateBlockStakeTransactionHandler handles API calls to POST /wallet/create/blockstaketransaction
+func (api *API) walletCreateBlockStakeTransactionHandler(w http.ResponseWriter, req *http.Request, _ httprouter.Params) {
+	var body WalletCreateBlockStakeTransactionPOST
+	if err := json.NewDecoder(req.Body).Decode(&body); err != nil {
+		WriteError(w, Error{"error decoding the supplied coin inputs and outputs: " + err.Error()}, http.StatusBadRequest)
```
GlenDC: `block inputs and outputs`*

api/wallet.go
```diff
@@ -630,9 +632,22 @@ func (api *API) walletCreateCoinTransactionHandler(w http.ResponseWriter, req *h
...
+		return
+	}
+	tx, err := api.wallet.CreateRawTransaction(nil, body.Inputs, nil, body.Outputs, nil)
+	if err != nil {
+		WriteError(w, Error{"error after call to /wallet/create/cointransaction: " + err.Error()}, http.StatusBadRequest)
```
GlenDC: `/wallet/create/blockstaketransaction`

modules/wallet/transactionbuilder.go
```diff
@@ -415,132 +415,130 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
-			}
-		case *types.UnlockHashCondition:
-			if key, exists := tb.wallet.keys[condition.UnlockHash()]; exists {
+// signBlockStakeInput attempts to sign a blockstake input with a key from the wallet
+func (tb *transactionBuilder) signBlockStakeInput(idx int, bsi *types.BlockStakeInput, cond types.MarshalableUnlockCondition) error {
+	switch condition := cond.(type) {
+	case *types.MultiSignatureCondition:
+		for _, uh := range condition.UnlockHashes {
+			if key, exists := tb.wallet.keys[uh]; exists {
 				if bsi.Fulfillment.Fulfillment == nil || bsi.Fulfillment.Fulfillment.Equal(nil) {
```
GlenDC: A simpler check would be `bsi.Fulfillment.FulfillmentType() == FulfillmentTypeNil`.

pkg/client/walletcmd.go
```diff
@@ -468,7 +479,7 @@ type outputPair struct {
 	Value     types.Currency
 }
 
-func parsePairedOutputs(args []string) (pairs []outputPair, err error) {
+func parsePairedOutputs(args []string, blockstake bool) (pairs []outputPair, err error) {
```
GlenDC: personally I would give it a functor `func(string) (types.Currency, error)` in a case like this, would avoid the weird if checking every iterator.

The `ParseCoinString` method already fits, the std ParseUint`, could be wrapped by a private function to do so as well.

However, it's a CLI tool, so I'm less worried about it here, even though imho it does make stuff a bit confusing, as such boolean check makes me always think that it might not be true for all iterations, hence why a boolean was chosen for this.

pkg/client/walletcmd.go
```diff
@@ -808,13 +822,57 @@ func walletcreatecointxn(cmd *cobra.Command, args []string) {
 	if err != nil {
 		Die("Could not create raw transaction from inputs and outputs: ", err)
 	}
-	var resp api.WalletCreateCoinTransactionRESP
```
GlenDC: Don't do that, please do keep it wrapped in a response structured. It might seem silly, but it gives the option in the future to add some extra property to the response, without breaking anything, while that isn't easily possible when returning a pure value.

pkg/client/walletcmd.go
```diff
@@ -808,13 +822,57 @@ func walletcreatecointxn(cmd *cobra.Command, args []string) {
...
+	var resp types.Transaction
 	err = _DefaultClient.httpClient.PostResp("/wallet/create/cointransaction", buffer.String(), &resp)
 	if err != nil {
 		Die("Failed to create transaction:", err)
 	}
 
-	json.NewEncoder(os.Stdout).Encode(resp.Transaction)
+	json.NewEncoder(os.Stdout).Encode(resp)
```
GlenDC: This error shouldn't be ignored I think.

pkg/client/walletcmd.go
```diff
@@ -808,13 +822,57 @@ func walletcreatecointxn(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		Die("Could not create raw transaction from inputs and outputs: ", err)
+	}
+	var resp types.Transaction
+	err = _DefaultClient.httpClient.PostResp("/wallet/create/blockstaketransaction", buffer.String(), &resp)
+	if err != nil {
+		Die("Failed to create transaction:", err)
+	}
+
+	json.NewEncoder(os.Stdout).Encode(resp)
```
GlenDC: This error shouldn't be ignored I think.

GlenDC:
@leesmet: I already took part of your code and merged it into master, as to already be able to get the unspent outputs from the consensus. Please rebase your commits so that your branch has no conflicts.

The commit, which contains some of your original code, can be seen at master on https://github.com/rivine/rivine/commit/39def46a633351fda132a2e8b12b7c079f73bf7a

pkg/client/walletcmd.go
LeeSmet: At this point the response has been encoded in the daemon endpoint, and decoded again by the cli,  so an error here would likely be caused by an error in the underlying `Write` call to `stdout`. So can we do something useful with this error then?

LeeSmet:
Rebased and updated, minus the "ignored error" of the json encoding in the cli, as I am not sure if we can do anything useful with the error should it occur. The json has already been encoded and decoded, so an error is likely form the underlying `Write` call, which would indicate a problem with writing to the output pipe.

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,134 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
+				Key:         key.SecretKey,
+			})
+			if err != nil {
+				return err
+			}
+			tb.signed = true
+		}
+	case *types.TimeLockCondition:
+		tb.signCoinInput(idx, ci, condition.Condition)
+	default:
```
GlenDC: You are not supporting *types.NilCondition` yet, while this can also be automatically signed.

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,134 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
+				Key:         key.SecretKey,
+			})
+			if err != nil {
+				return err
+			}
+			tb.signed = true
+		}
+	case *types.TimeLockCondition:
+		tb.signBlockStakeInput(idx, bsi, condition.Condition)
+	default:
```
GlenDC: You are not supporting *types.NilCondition` yet, while this can also be automatically signed.

pkg/client/walletcmd.go
```diff
@@ -430,7 +479,19 @@ type outputPair struct {
...
 
-func parsePairedOutputs(args []string) (pairs []outputPair, err error) {
+// parseCurrencyString takes the string representation of a currency value
+type parseCurrencyString func(string) (types.Currency, error)
+
+func stringToCoins(input string) (types.Currency, error) {
+	return _CurrencyConvertor.ParseCoinString(input)
+}
+
+func stringToBlockStake(input string) (types.Currency, error) {
```
GlenDC: `stringToBlockstakes`*

pkg/client/walletcmd.go
```diff
@@ -430,7 +479,19 @@ type outputPair struct {
 	Value     types.Currency
 }
 
-func parsePairedOutputs(args []string) (pairs []outputPair, err error) {
+// parseCurrencyString takes the string representation of a currency value
+type parseCurrencyString func(string) (types.Currency, error)
+
+func stringToCoins(input string) (types.Currency, error) {
```
GlenDC: No need for this function, yo can just use `_CurrencyConvertor.ParseCoinString` directly.

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,174 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
+				Transaction: tb.transaction,
+				Key:         key.SecretKey,
+			})
+			if err != nil {
+				return err
+			}
+			tb.signed = true
+		}
+	case *types.TimeLockCondition:
+		tb.signCoinInput(idx, ci, condition.Condition)
```
GlenDC: You forgot a `return` keyword here, no?

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,174 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
+				Transaction: tb.transaction,
+				Key:         key.SecretKey,
+			})
+			if err != nil {
+				return err
+			}
+			tb.signed = true
+		}
+	case *types.TimeLockCondition:
+		tb.signBlockStakeInput(idx, bsi, condition.Condition)
```
GlenDC: You forgot a `return` keyword here, no?

modules/wallet/transactionbuilder.go
```diff
@@ -413,6 +414,174 @@ func (tb *transactionBuilder) Sign() ([]types.Transaction, error) {
...
+	case *types.NilCondition:
+		// Sign using a random key, if any is present
+		if len(tb.wallet.keys) > 0 {
+			var key spendableKey
+			for _, key = range tb.wallet.keys {
```
GlenDC: This doesn't get a crypto-random key though. Either get a crypto-random key, or use perhaps just `getAddress`, which is meant for those purposes.

PR #368, opened 2018-05-31
Close #337: Merge transactions

### LINKED ISSUE #337: add `rivinec merge inputs` command

LeeSmet:


GlenDC:
Unit tests and some manual end-to-end test seem to indicate that this feature works as is, so merging it.

Issue #366, opened 2018-05-30
support JSON outputs of atomic swap commands


GlenDC:
We can achieve this by:

1. adding a persistent flag (`--encoding`) to the `atomicSwapCmd` using https://godoc.org/github.com/spf13/cobra#Command.PersistentFlags, and calling the `Var` method on that returned flag set (see the explore.go file and the consensus's transaction CLI command of examples on how to register a `cli.EncodingType` as a `EncodingTypeFlag` using that `Var` method, as mask you'll want to use `EncodingTypeHuman | EncodingTypeJSON`, as to not allow hex encoding, given it doesn't make any sense here)
2. defining structures with JSON annotation tags for each output type, as a public type, so these are included in the documentation, you'll need one for each command
3. adapt the command methods, such that where we output data to the STDOUT, we first switch on the `atomicSwapcfg.EncodingType` property, prior to printing. If it equal the `EncodingTypeHuman`, we would print in the human STDOUT format we currently use, a very custom but readable one, but if it's `EncodingTypeJSON`, we would print using the correctly defined struct, defined for this JSON-encoded printing purpose only. You'll need to do this in the following places:
    1. outputID+transactionID of created contract @ https://github.com/rivine/rivine/blob/a4d74b1139ee4f0918c74d59a7d919a73f086323/pkg/client/atomicswap.go#L251
    2. audited contract's properties and value @ https://github.com/rivine/rivine/blob/a4d74b1139ee4f0918c74d59a7d919a73f086323/pkg/client/atomicswap.go#L341 (and print those optional validation error messages below in the STDERR instead of STDOUT, no matter the encoding format, so that STDOUT/STDERR separation is properly done there)
    3. secret, extracted from contract @ https://github.com/rivine/rivine/blob/a4d74b1139ee4f0918c74d59a7d919a73f086323/pkg/client/atomicswap.go#L496
    4. transactionID from transaction used to redeem/refund  contract @ https://github.com/rivine/rivine/blob/a4d74b1139ee4f0918c74d59a7d919a73f086323/pkg/client/atomicswap.go#L636

This issue has to be resolved should you want to use the atomic swap commands in an automated environment. Should you resolve this issue together with #365, you'll be able to use the atomic swap commands in an automated environment.

jdelrue:
GlenDC:

The way I solved it in the current version:
- Added a printer.go file that contains human readable and json name for each printerproperty
https://github.com/JimberSoftware/rivine/blob/master/pkg/client/printer.go

[...]

Issue #365, opened 2018-05-30
support `-y/--yes` global flag to the atomicSwapCmd


GlenDC:
When this (boolean) flag is defined as `true`, it should prevent any Y/N questions being asked to the user in the atomic swap commands.

The global flag can be added to `atomicSwapCmd ` using https://godoc.org/github.com/spf13/cobra#Command.PersistentFlags, and calling the BoolVarP method on that returned flag set (see other flags in our client code to see our conventions about how to define these flag properties and such).

Once this flag is added we want to put an `if !atomicSwapcfg.YesToAll {}` check around following code blocks: 

+ https://github.com/rivine/rivine/blob/a4d74b1139ee4f0918c74d59a7d919a73f086323/pkg/client/atomicswap.go#L217-L224
+ https://github.com/rivine/rivine/blob/a4d74b1139ee4f0918c74d59a7d919a73f086323/pkg/client/atomicswap.go#L590-L597

This check will ensure two things:

1. To not print the contract to STDOUT, and thus not interfering with the regular output, a very important thing for an automated environment;
2. To not ask a Y/N question and simply assume Yes instead for this question, as indicated by the `-y/--yes` flag;

This issue has to be resolved should you want to use the atomic swap commands in an automated environment.

jdelrue:
Code for this is ready. Will do pull request today!

jdelrue:
Why don't we adjust

``
func askYesNoQuestion(str string) bool {
``

so it returns the value? This way we don't have to include a check in every individual code block.

If we don't print to std::out, how will we ever interpret (json) output in an automated environment?

jdelrue:
Impl ready and pushed, will test this afternoon

GlenDC:
You proposed earlier to put the `YesToAll` property check in the `askYesNoQuestion ` function (https://github.com/rivine/rivine/blob/84349693e539652219bbb7d94f911244c4ffbec4/pkg/client/atomicswap.go#L716), at the top, and yes I agree with that. Feel free to put an `if yesToAll { return true }` check at the top.

I thought about this before, but at that point it made sense to me to not print the contract review. However, that does mean that you wan to ensure that https://github.com/rivine/rivine/blob/84349693e539652219bbb7d94f911244c4ffbec4/pkg/client/atomicswap.go#L218-L219 is only executed when the encodingType is `human` (a check possible for when #366 is (being) resolved).

[...]

PR #372, opened 2018-06-01
Implemented #366 #365 encoding and yes flag

jdelrue:


GlenDC:
Rebased your commits and combined it with another commit in a new PR (https://github.com/rivine/rivine/pull/373), so will close this one.

PR #375, opened 2018-06-04
provide new optional function for daemon config

GlenDC:
this function allows you to register/unregister unlock conditions/fulfillments
custom to your chain (network)

also fixing a typo in an already existing function within that config


GlenDC:
This function is overkill, keeping it as part of the network config makes sense.

PR #388, opened 2018-06-07
Fix some locking issues

LeeSmet:


modules/gateway/ip.go
```diff
@@ -37,9 +37,7 @@ func (g *Gateway) managedIPFromPeers() (string, error) {
 		default:
 		}
 		// Get peers
-		g.mu.RLock()
```
GlenDC: Can you please comment why the removal of these locks is OK?

FYI, these locks are still there in Sia: https://github.com/NebulousLabs/Sia/blob/dfebf75ba2425125666dd8327e224be06b272d77/modules/gateway/ip.go#L39-L42

LeeSmet: The function they were guarding starts by acquiring an `RLock`, and defers the `RUnlock`. So the same lock is just grabbed twice

GlenDC: https://github.com/NebulousLabs/Sia/pull/3085

PR #431, opened 2018-08-05
Electrum

### LINKED ISSUE #408: support a 'notificator' module which allows you to subscribe to certain events
Initially the only event we would support is changes to a given address (unlockhash), as is requested by the motivating User-Facing issue: threefoldfoundation/tfchain#145

Through websockets we should expose a new protocol using JSON-RPC, which allows users to subscribe to a certain event "stream" and receive notifications for the desired subscriptions when needed. This to prevent the need for a client to having to randomly poll the server, generating traffic for no good reason, and instead receive the desired updates when they happen, without much overhead.

Subscribing to an address will be the first use-case, such that a light client could subscribe to a the address(es) it owns, and receive (un)spend outputs as they happen.

### LINKED ISSUE #432: Electrum: write unit tests

LeeSmet:
Add basic electrum support. The only functional call is `blockchain.address.subscribe`, which will push updates to the client if the address status changes. For this purpose address is calculated as introduced in the original [electrum spec](https://electrumx.readthedocs.io/en/latest/protocol-basics.html#status).

Two more calls are implemented which serve the protocol itself, being `server.ping`, a utility method for keeping the connection alive, and `server.version`, a method for negotiating the protocol version. Right now we only support `1.0.0`, and clients must mark the protocol version as exactly this or they will be disconnected (there is no version negotiation implemented yet). Regardless, the call itself is already in place to make sure clients already implement it as well, to have it ready for future versions (and to make sure it is not a breaking change if we would omit it now and suddenly require it later).

Closes #408
Closes #432


modules/electrum/errors.go
```diff
@@ -0,0 +1,36 @@
...
+	ErrInvalidParams = RPCError{
+		Code:    -32602,
+		Message: "Invalid parameters",
+	}
+	ErrInternal = RPCError{
+		Code:    -32603,
+		Message: "Internal error",
+	}
+)
+
```
GlenDC: if you use any of the `32000 to -32099` server error code range, best to define them here as a separate var list as well, so that it is clear in our public API.

modules/electrum/protocolversion.go
```diff
@@ -0,0 +1,117 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type (
+
```
GlenDC: Another empty line.

modules/electrum/protocolversion.go
```diff
@@ -0,0 +1,117 @@
...
+	// ProtocolVersion is a specific version of the electrum protocol.
+	// The version is denoted by a major and minor number, and optionally, a rivision
+	ProtocolVersion struct {
```
GlenDC: You miss a `Cmp` method (or alternative approach), that you'll need as part of the server negotiation.

modules/electrum/protocolversion.go
```diff
@@ -0,0 +1,117 @@
...
+		major    uint8
+		minor    uint8
+		revision uint8
+	}
+
+	// ProtocolArgument is the input for the `server.version` call's `protocol_version`
+	// argument. As per the spec, it is either a single version, or an array
+	// [protocol_min, protocol_max].
+	ProtocolArgument struct {
```
GlenDC: Why is this defined here, shouldn't it be defined as part of the Response/Request structs of the RPC call in question?

modules/electrum/protocolversion.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := epv.ParseRawVersion(rawVersion); err != nil {
+		return err
+	}
+	return nil
+}
+
+// ParseRawVersion tries to parse a version string. If parsing succeeds (no error
+// is returned), the version information is updated.
+func (epv *ProtocolVersion) ParseRawVersion(rawVersion string) error {
+	segments := strings.Split(rawVersion, ".")
```
GlenDC: You can simplify your code here to just a couple of lines:

```go
n, err := fmt.Sscanf(rawVersion + ".0", "%d.%d.%d", &epv.major, &epv.minor, &epv.revision)
if err != nil {
    return err
}
if n != 3 { // due to our added `+"0"` we should always have at least 3 segments
    return errors.New("Invalid amount of segments")
}
return nil
```

modules/electrum/rpc.go
```diff
@@ -0,0 +1,276 @@
...
+		error
+	}
+
+	// Request is the structure of an rpc call in JSONRPC 2.0. The JSONRPC field
+	// value must be exactly "2.0", any other value, or its absence, is an error
+	Request struct {
+		JSONRPC string           `json:"jsonrpc"`
+		Method  string           `json:"method"`
+		Params  *json.RawMessage `json:"params"`
+		ID      interface{}      `json:"id"`
```
GlenDC: As far as I understand, ID cannot be a boolean or an object. It can only be a `string`, `null` or a `number`. Defining the ID like this allows you to indeed accept all 3 valid cases, but unless I miss something, I am fairly sure that you currently would also accept booleans and objects.

GlenDC: Shouldn't it be `json:"id,omitempty"` by the way, given that the request can omit the ID in case it does not care about a response? Or Do I misunderstand something here?

modules/electrum/rpc.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	r.Result = result
+	return r
+}
+
+// ServeRPC serves the electrum protocol on the given transport.
+// This blocks untill the connection is closed
+func (e *Electrum) ServeRPC(transport RPCTransport) {
+	e.activeConnections.Add(1)
```
GlenDC: Wouldn't it be safer to `defer` the `e.activeConnections.Done()` call, such that it will be always run when the `ServeRPC` is finished, as well as be more clear. Or is there a good reason for this current disconnect?

modules/electrum/rpc.go
```diff
@@ -0,0 +1,276 @@
...
+
+	cl := &Client{
+		transport:            transport,
+		timer:                time.NewTimer(connectionTimeout),
+		serviceMap:           make(map[string]rpcFunc),
+		addressSubscriptions: make(map[types.UnlockHash]bool),
```
GlenDC: why `bool` and not `struct{}`? The first takes value memory, while the latter does not.

modules/electrum/rpc.go
```diff
@@ -0,0 +1,276 @@
...
+
+			e.log.Debugln("Client error on connection:", err)
+			// If we are here the connection is alive but the request could
+			// not be parsed. Send a parse error
+			response := &Response{
+				Error:   &ErrParse,
+				JSONRPC: jsonRPCVersion,
+			}
+			cl.transport.Send(response)
+			cl.resetTimer()
```
GlenDC: Why the reset timer here?

modules/electrum/rpc.go
```diff
@@ -0,0 +1,276 @@
...
+			}
+
+			e.log.Debugln("Client error on connection:", err)
+			// If we are here the connection is alive but the request could
+			// not be parsed. Send a parse error
+			response := &Response{
+				Error:   &ErrParse,
+				JSONRPC: jsonRPCVersion,
+			}
+			cl.transport.Send(response)
```
GlenDC: error result is ignored here

modules/electrum/rpc.go
```diff
@@ -0,0 +1,276 @@
...
+					cl.wg.Done()
+					e.log.Println("Closing connection to", cl.transport.RemoteAddr(), "due to an error")
+					e.closeConnection(cl)
+					return
+				default:
+				}
+
+				// Avoid sending nil
+				if resp.MustSend() {
+					cl.transport.Send(resp)
```
GlenDC: error result is ignored here

modules/electrum/tcp.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (c *tcpConn) GetRequest() <-chan *BatchRequest {
+	return c.requestChan
+}
+
+func (c *tcpConn) Send(msg interface{}) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.enc.Encode(msg)
```
GlenDC: wouldn't this be cheaper —and just as safe— if done over a channel?

modules/electrum/update.go
```diff
@@ -0,0 +1,45 @@
...
+	// Update is a collection of info processed from consensus set/
+	// explorer/transactionpool updates
+	Update struct {
+		addressStates map[types.UnlockHash]string
+	}
+)
+
+// ProcessConsensusChange parses updates from the cs to a general format,
+// which is then passed on to all connections.
+func (e *Electrum) ProcessConsensusChange(cc modules.ConsensusChange) {
```
GlenDC: This seems like a very expensive loop:

* An address state computation might be done multiple times, if defined multiple times;
* A state computation means it goes through all outputs and inputs of the address, meaning this makes every state even more expensive;

Having to this every block again and again (perhaps multiple times) makes this a very expensive call. Especially given that we currently have a couple of addresses with a lot of outputs linked to them.

modules/electrum/update.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	for _, bsod := range cc.BlockStakeOutputDiffs {
+		address := bsod.BlockStakeOutput.Condition.UnlockHash()
+		// avoid computing status for an address multiple times
+		if _, exists := change.addressStates[address]; !exists {
+			change.addressStates[address] = e.AddressStatus(address)
+		}
+	}
+
```
GlenDC: Don't we need to loop through inputs as well. If an input is used, it means the state of the address (Which owned that input) has been changed as well, no? (e.g. decrease of balance).

modules/electrum/websocket.go
```diff
@@ -0,0 +1,113 @@
...
+	requestChan chan *BatchRequest
+	errorChan   chan error
+	stopChan    chan struct{}
+
+	// Synchronize writes
+	mu sync.Mutex
+}
+
+var upgrader = websocket.Upgrader{
+	ReadBufferSize:  1024,
```
GlenDC: please comment the origin of these numbers

modules/electrum/update.go
LeeSmet: The `ConsensusChange` struct does not have inputs, it only has `diffs` of unspent outputs. If an output is used (so if it is an input), it will be a diff where the used output is removed. So looping over *all* output diffs, getting the address of the output, should suffice imo

modules/electrum/update.go
LeeSmet: Only the addresses which change in a block are computed. The status string is used as a value in  a map with the address as key, and we check if the key is present before calculating. So this should only calculate the status string once per address?

modules/electrum/tcp.go
LeeSmet: I suppose you would then send a possible error over the error channel?

modules/electrum/rpc.go
LeeSmet: If we are here we encountered a parse error, which means the client did sent something (though incorrectly formatted). But the client is there nonetheless so keep the connection open.

modules/electrum/client.go
```diff
@@ -0,0 +1,102 @@
...
+		protoVersion ProtocolVersion
+	}
+)
+
+func (cl *Client) registerService(name string, f rpcFunc) error {
+	cl.mu.Lock()
+	defer cl.mu.Unlock()
+
+	if _, exists := cl.serviceMap[name]; exists {
+		return errors.New("A function with name " + name + " already exists for this connection")
```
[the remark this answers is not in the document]
LeeSmet: Yes, it should never happen. If it does, the error returned will get logged and the calling function will return an internal server error

LeeSmet:
The version negotiation has not been implemented properly as we decided to not do it to save some time since we only have one version for now. So I think it is acceptable to only accept v1.0 as max version (as its the only we have). For now the code is just in place so we already have the API as it should be when completely finished

modules/electrum/websocket.go
```diff
@@ -20,6 +20,10 @@ type websocketConn struct {
 }
 
 var upgrader = websocket.Upgrader{
+	// ReadBufferSize and WriteBufferSize specify the sizes of the buffers,
```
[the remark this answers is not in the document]
LeeSmet: Not really, the only reason I picked these specifically is because the library which is used uses the same values in their API documentation, so I am assuming these are sane to use defaults.

modules/electrum/errors.go
```diff
@@ -2,27 +2,45 @@ package electrum
...
+	ErrCodeMethodNotFound       = -32601
+	ErrCodeInvalidParams        = -32602
+	ErrCodeInternalError        = -32603
+)
+
+// Error codes we define ourselves. These error codes
+// indicate an error which does not map cleanely to the
+// predefined JSONRPC errors.
+const (
+	ErrCodeVersionAlreadySet = 101
```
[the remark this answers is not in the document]
LeeSmet: Iirc I was looking at another lib the other day which started their error codes at 101 as well (they reserved the first 99 errors for some reason). And it seems weird to have negative error codes to me. But there really is no good reason

doc/examples/electrum/websocket.html
```diff
@@ -5,11 +5,22 @@
...
 
   function init()
   {
     output = document.getElementById("output");
     testWebSocket();
+
+    // Keep the connection open by sending keep alive pings.
+    // Even though the server disconnects after 10 minutes, the pings
+    // are send every 9 minutes. This ensures we are not disconnected because
+    // of network lag. The ping handler could optionally be expanded to
```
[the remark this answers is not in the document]
LeeSmet: added some info about this in the electrum README

LeeSmet:
The error from the example is because the websocket port was not enabled when starting the deamon.  You need to manually enable the tcp and or ws ports using flags. This is documented in the electrum README

LeeSmet:
I am going to merge this PR into the development branch now as discussed with @robvanmieghem. Any other comments / changes can be applied there

PR #449, opened 2018-08-29
stop IBD if one OB peer is synced and no blocks are incoming

### LINKED ISSUE #447: consensus stuck in threadedInitialBlockchainDownload if majority of outbound peers are out of sync
Currently our consensus module is stuck in an infinite loop within the `threadedInitialBlockchainDownload` function, as long as we don't have more outbound peers synced than we have not. This is normally not an issue, but if for some kind of upgrade, there is a network with not many users, who all upgrade at the same time, you might get in the situation where not enough are in sync.

There is a max deadline, but it is only used when more than half of the outbound peers are in sync. So not much of a help here, see: https://github.com/rivine/rivine/blob/f44a94b7d0ffc8eb076cf5e9fc0591a74297ac15/modules/consensus/synchronize.go#L573

I guess there are a couple of approaches we can take here:

* We could kick out out of sync nodes, hoping other nodes could connect that are in sync, but not sure if that would help and it might still give issues;
* We could use that deadline for all situations, and just mark as sync after the deadline regardless (as long as we have peers connected);
* See if there is something to fix in that logic, as the height doesn't seem to be taken into account, I would assume that an outbound peer could also count in sync if it has the same height as the marked-in-sync outbound peers (not sure if this is possible in the current RPC protocol though);

Whatever we decide to do, something must be done, as right now that line is causing issues in the tfchain testnet nodes. Given that Threefoldfoundation for now still controls most of the nodes on testnet, meaning they all upgrade at once, and thus don't get to a majority of marked-in-sync outbound peers.

GlenDC:
closes #447

### DIFF
--- a/modules/consensus/synchronize.go
+++ b/modules/consensus/synchronize.go
@@ -50,23 +50,6 @@ var (
 			panic("unrecognized build.Release")
 		}
 	}()
-	// minIBDWaitTime is the time threadedInitialBlockchainDownload waits before
-	// exiting if there are >= 1 and <= minNumOutbound peers synced. This timeout
-	// will primarily affect people who have multiple nodes daisy chained off each
-	// other. Those nodes will likely have to wait minIBDWaitTime on every startup
-	// before IBD is done.
-	minIBDWaitTime = func() time.Duration {
-		switch build.Release {
-		case "dev":
-			return 80 * time.Second
-		case "standard":
-			return 90 * time.Minute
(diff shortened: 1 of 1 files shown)


GlenDC:
Required for `tfchain` testnet, as that is currently going since the last 24 hours on a slow pace. You can still review whenever you can/want @robvanmieghem, but it will be when already merged into master.

PR #462, opened 2018-11-13
Erc20 demo

### LINKED ISSUE #458: Investigate how to monitor erc20 tokens for a erc20 bridge
Create a small example:

- Contract on erc20 testnet which has some erc20 tokens (see erc20 wiki for example contract).
- Server which monitors said contract. This server is effectively a light client.
- A small webpage which will talk to our own server and shows some info (can be made to look like an exchange?)

Code can be put in the doc folder, as it can be used later in demo's.

For the webpage it might be interesting to use the latest angular.

LeeSmet:
Example UI which monitors a single address of the TTFT20 contract on the rinkeby testnet.

Close #458


LeeSmet:
Added erc20 oracle prototype

Issue #490, opened 2019-03-07
integrate parseDescription in Rivine CLI Code


GlenDC:
TFChain added code in their codebase to parse the description passed to the CLI (`create`) commands, such that not only a regular string can be passed, but also escaped characters can be passed (e.g. binary or unicode). Goal of this issue is to move that logic over to Rivine, e.g. make a new flag function for it, and use it also for all description flags (usually called `data`) in Rivine CLI commands.

The relevant TFChain commits, so you know what to move over (not a plane copy, but more of a migration, as we'll want it a bit different):
* https://github.com/threefoldfoundation/tfchain/commit/cea2e5927482ace23956c8ad2eadf46b1929535c
* https://github.com/threefoldfoundation/tfchain/commit/eeb366342f44fcf1d816fa871f68d32ae77e39b4
* https://github.com/threefoldfoundation/tfchain/commit/98c38a6f89de618f5885ae0674864eae26b7136b

Tasks
* [ ] copy the `parseDescription` logic and make a [pflag.Value implementation](https://godoc.org/github.com/spf13/pflag#Value) of it, and add that logic to `flag.go` and `flag_test.go` in https://github.com/threefoldtech/rivine/blob/master/pkg/cli, the `Value` implementation, calling it `arbitraryData`, will be similarly implemented as can be seen in https://github.com/threefoldfoundation/tfchain/blob/98c38a6f89de618f5885ae0674864eae26b7136b/pkg/cli/flag.go#L10-L59 (the underlying type will be a `[]byte`
* [ ] ensure to add unit tests for this added flag
* [ ] use this flag constructor everywhere in the commands of the Rivine CLI code where we have a `--data` or `--description` flag
* [ ] make a PR

PR #493, opened 2019-03-11
Parse description of (--data) flag cli

DylanVerstraete:
Fixes issue #490


pkg/cli/flag.go
```diff
@@ -219,6 +223,50 @@ func EncodingTypeFlagDescription(mask EncodingType) string {
 	return "enum flag to define how to encode the output, options: " + strings.Join(options, "|")
 }
 
+// ArbitraryDataFlagVar defines a []modules.NetAddress flag with specified name and usage string.
```
GlenDC: Copy-pasting is OK, but do make sure the content is up-to-date, in this case, update comments

pkg/cli/flag.go
```diff
@@ -219,6 +223,50 @@ func EncodingTypeFlagDescription(mask EncodingType) string {
...
+// The argument s points to a []modules.NetAddress variable in which to store the validated values of the flags.
+// The value of each argument will not try to be separated by comma, each value has to be defined as a separate flag (using the same name).
+func ArbitraryDataFlagVar(f *pflag.FlagSet, s *[]byte, name string, usage string) {
+	f.Var(&ArbitraryDataFlag{arbitraryData: s}, name, usage)
+}
+
+// ArbitraryDataFlagVarP defines a []modules.NetAddress flag with specified name, shorthand and usage string.
```
GlenDC: Copy-pasting is OK, but do make sure the content is up-to-date, in this case, update comments

pkg/cli/flag.go
```diff
@@ -219,6 +223,50 @@ func EncodingTypeFlagDescription(mask EncodingType) string {
...
+// String implements pflag.Value.String
+func (flag *ArbitraryDataFlag) String() string {
+	byteArray := *flag.arbitraryData
+	if len(byteArray) == 0 {
+		return ""
+	}
+
+	s := strconv.Quote(string(byteArray[:]))
+
+	if s == "" {
```
GlenDC: this can never happen, as you already check `len(byteArray)`

pkg/cli/flag.go
```diff
@@ -219,6 +223,50 @@ func EncodingTypeFlagDescription(mask EncodingType) string {
...
+		return ""
+	}
+	return strings.Trim(s, "\"")
```
GlenDC: Please check docs, to make sure what quotes it uses, is it always `"`, or can it also be another one? If so, you might need to check which one is used and trim that one.

pkg/cli/flag_test.go
```diff
@@ -302,6 +303,44 @@ func testPanic(t *testing.T, label string, f func()) {
...
+			[]byte{0, 1, 2},
+		}, {
+			"2",
+			`\x00\x05\x13`,
+			[]byte{0, 5, 19},
+		}, {
+			"3",
+			`\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00`,
+			[]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0},
+		},
```
GlenDC: Also check edge cases, in your case, the empty string. Plus also check regular strings, e.g. `"Hello, World!"`, as well as a string such as a unicode string escaped AND non-escaped.

pkg/client/walletcmd.go
```diff
@@ -296,6 +291,12 @@ func createWalletCmd(cli *CommandLineClient) *WalletCommand {
 		&walletCmd.walletLoadSeedCfg.Seed,
 		"seed", "", "define the seed to be loaded as a flag instead of the STDIN")
 
+	// custom arbitrarydata flag
+	clipkg.ArbitraryDataFlagVar(sendCoinsCmd.Flags(), &walletCmd.sendCoinsCfg.Data,
+		"data", "optional arbitrary data (or description) to attach to transaction")
+	clipkg.ArbitraryDataFlagVar(sendCoinsCmd.Flags(), &walletCmd.sendBlockStakesCfg.Data,
```
GlenDC: copy-paste error

pkg/cli/flag.go
DylanVerstraete: https://golang.org/src/strconv/quote.go?s=2985:3012#L104

`Quote returns a double-quoted Go string literal`

GlenDC:
LGTM

PR #498, opened 2019-03-11
refactored panic handling - using build severe

### LINKED ISSUE #489: Clean up if build.DEBUG ..panic everywhere
At a lot of places some code like 
```
if build.DEBUG && err != nil {
			panic(err)
}
```
exists, it should be replaced with a call to 
```
import build
build.Severe(
```
or `build.Critical`
As those  ones panic on `build.DEBUG`as well and always log the error while now it is often neglected.

DylanVerstraete:
Fixes #489 

Using `build.Severe` right now almost everywhere. Maybe there is some indication of where I can use these `build.Critical` statements? Tests required some refactor also.

### DIFF
--- a/cmd/rivinec/main.go
+++ b/cmd/rivinec/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/threefoldtech/rivine/build"
 	"github.com/threefoldtech/rivine/modules"
 	"github.com/threefoldtech/rivine/pkg/cli"
 	"github.com/threefoldtech/rivine/pkg/client"
@@ -16,7 +17,7 @@ func main() {
 	bchainInfo := types.DefaultBlockchainInfo()
 	cliClient, err := client.NewCommandLineClient("", bchainInfo.Name, daemon.RivineUserAgent)
 	if err != nil {
-		panic(err)
+		build.Severe(err)
 	}
 	// define preRunE, as to ensure we go to a default config should it be required
 	cliClient.PreRunE = func(cfg *client.Config) (*client.Config, error) {
--- a/cmd/rivinec/main.go
+++ b/cmd/rivinec/main.go
@@ -17,7 +17,7 @@ func main() {
 	bchainInfo := types.DefaultBlockchainInfo()
 	cliClient, err := client.NewCommandLineClient("", bchainInfo.Name, daemon.RivineUserAgent)
 	if err != nil {
-		build.Severe(err)
+		build.Critical(err)
 	}
 	// define preRunE, as to ensure we go to a default config should it be required
 	cliClient.PreRunE = func(cfg *client.Config) (*client.Config, error) {
--- a/cmd/rivined/daemon.go
+++ b/cmd/rivined/daemon.go
@@ -183,7 +183,7 @@ func runDaemon(cfg daemon.Config, networkCfg daemon.NetworkConfig, moduleIdentif
 		// need to flush the response before shutting down the server
 		f, ok := w.(http.Flusher)
 		if !ok {
-			err := errors.New("cannot revert to block outside of current path")
+			err := errors.New("Server does not support flushing")
 			build.Severe(err)
 		}
 		f.Flush()
--- a/cmd/rivined/daemon.go
+++ b/cmd/rivined/daemon.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -9,6 +10,7 @@ import (
 	"time"
 
 	"github.com/julienschmidt/httprouter"
+	"github.com/threefoldtech/rivine/build"
 	"github.com/threefoldtech/rivine/modules"
 	"github.com/threefoldtech/rivine/modules/blockcreator"
 	"github.com/threefoldtech/rivine/modules/consensus"
@@ -181,7 +183,8 @@ func runDaemon(cfg daemon.Config, networkCfg daemon.NetworkConfig, moduleIdentif
 		// need to flush the response before shutting down the server
 		f, ok := w.(http.Flusher)
 		if !ok {
-			panic("Server does not support flushing")
+			err := errors.New("cannot revert to block outside of current path")
+			build.Severe(err)
 		}
 		f.Flush()
 
--- a/crypto/signatures_test.go
+++ b/crypto/signatures_test.go
@@ -2,6 +2,7 @@ package crypto
 
 import (
 	"bytes"
+	"strings"
 	"testing"
 
 	"github.com/NebulousLabs/fastrand"
@@ -171,7 +172,9 @@ func TestReadWriteSignedObject(t *testing.T) {
 
 	// Disrupt the decoding to get coverage on the failure branch.
 	err = ReadSignedObject(b, &read, 11, pk)
-	if err == nil || err.Error() != "could not decode type crypto.Signature: unexpected EOF" {
+	contains := strings.Contains(err.Error(), "could not decode type crypto.Signature: Severe")
+
+	if err == nil || !contains {
 		t.Fatal(err)
 	}
 
--- a/crypto/signatures_test.go
+++ b/crypto/signatures_test.go
@@ -172,7 +172,7 @@ func TestReadWriteSignedObject(t *testing.T) {
 
 	// Disrupt the decoding to get coverage on the failure branch.
 	err = ReadSignedObject(b, &read, 11, pk)
-	contains := strings.Contains(err.Error(), "could not decode type crypto.Signature: Severe")
+	contains := strings.Contains(err.Error(), "could not decode type crypto.Signature")
 
 	if err == nil || !contains {
 		t.Fatal(err)
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -4,11 +4,10 @@ import (
 	"errors"
 	"time"
 
+	bolt "github.com/rivine/bbolt"
 	"github.com/threefoldtech/rivine/build"
 	"github.com/threefoldtech/rivine/modules"
 	"github.com/threefoldtech/rivine/types"
-
-	"github.com/rivine/bbolt"
 )
 
 var (
@@ -135,8 +134,8 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
 	var nonExtending bool
 	err = cs.db.Update(func(tx *bolt.Tx) error {
 		pb, err := getBlockMap(tx, b.ParentID)
-		if build.DEBUG && err != nil {
-			panic(err)
+		if err != nil {
+			build.Critical(err)
 		}
 		currentNode := currentProcessedBlock(tx)
 		newNode := cs.newChild(tx, pb, b)
@@ -259,7 +258,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 	}
 	// If appliedBlocks is 0, revertedBlocks will also be 0.
 	if build.DEBUG && len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
-		panic("appliedBlocks and revertedBlocks are mismatched!")
+		build.Severe("appliedBlocks and revertedBlocks are mismatched!")
 	}
 	// Updates complete, demote the lock.
 	cs.mu.Demote()
--- a/modules/consensus/accept.go
+++ b/modules/consensus/accept.go
@@ -257,7 +257,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 		return err
 	}
 	// If appliedBlocks is 0, revertedBlocks will also be 0.
-	if build.DEBUG && len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
+	if len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
 		build.Severe("appliedBlocks and revertedBlocks are mismatched!")
 	}
 	// Updates complete, demote the lock.
--- a/modules/consensus/consensusdb.go
+++ b/modules/consensus/consensusdb.go
@@ -7,12 +7,11 @@ package consensus
 // ignored otherwise, which is suboptimal.
 
 import (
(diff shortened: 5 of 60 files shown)


cmd/rivinec/main.go
```diff
@@ -16,7 +17,7 @@ func main() {
 	bchainInfo := types.DefaultBlockchainInfo()
 	cliClient, err := client.NewCommandLineClient("", bchainInfo.Name, daemon.RivineUserAgent)
 	if err != nil {
-		panic(err)
+		build.Severe(err)
```
robvanmieghem: This would always have paniced, I would have used Critical here.

cmd/rivined/daemon.go
```diff
@@ -181,7 +183,8 @@ func runDaemon(cfg daemon.Config, networkCfg daemon.NetworkConfig, moduleIdentif
 		// need to flush the response before shutting down the server
 		f, ok := w.(http.Flusher)
 		if !ok {
-			panic("Server does not support flushing")
+			err := errors.New("cannot revert to block outside of current path")
```
robvanmieghem: Please use the adapt copy-pasted code to the correctt error

cmd/rivinec/main.go
GlenDC: That is not what build.Critical does though, I read the bodies of those 2 functions on Dylan's computer the other day, and they seemed to do the same.

crypto/signatures_test.go
```diff
@@ -171,7 +172,9 @@ func TestReadWriteSignedObject(t *testing.T) {
 
 	// Disrupt the decoding to get coverage on the failure branch.
 	err = ReadSignedObject(b, &read, 11, pk)
-	if err == nil || err.Error() != "could not decode type crypto.Signature: unexpected EOF" {
+	contains := strings.Contains(err.Error(), "could not decode type crypto.Signature: Severe")
```
robvanmieghem: what does this specific change have to do with the the refactoring?

modules/consensus/accept.go
```diff
@@ -259,7 +258,7 @@ func (cs *ConsensusSet) managedAcceptBlock(b types.Block) error {
 	}
 	// If appliedBlocks is 0, revertedBlocks will also be 0.
 	if build.DEBUG && len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
```
robvanmieghem: please refactor out the build.DEBUG:
```
if  len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {
build.Severe("appliedBlocks and revertedBlocks are mismatched!")
```

modules/consensus/consensusdb.go
```diff
@@ -272,12 +271,12 @@ func addCoinOutput(tx *bolt.Tx, id types.CoinOutputID, sco types.CoinOutput) {
 	*/
 	coinOutputs := tx.Bucket(CoinOutputs)
 	// Sanity check - should not be adding an item that exists.
-	if build.DEBUG && coinOutputs.Get(id[:]) != nil {
-		panic("repeat siacoin output")
+	if coinOutputs.Get(id[:]) != nil {
+		build.Severe("repeat siacoin output")
```
robvanmieghem: while at it, please modify to "repeat coin output"

modules/consensus/consensusdb.go
```diff
@@ -286,12 +285,12 @@ func addCoinOutput(tx *bolt.Tx, id types.CoinOutputID, sco types.CoinOutput) {
 func removeCoinOutput(tx *bolt.Tx, id types.CoinOutputID) {
 	scoBucket := tx.Bucket(CoinOutputs)
 	// Sanity check - should not be removing an item that is not in the db.
-	if build.DEBUG && scoBucket.Get(id[:]) == nil {
-		panic("nil siacoin output")
+	if scoBucket.Get(id[:]) == nil {
+		build.Severe("nil siacoin output")
```
robvanmieghem: while at it, please modify to "nil coin output"

modules/consensus/accept.go
robvanmieghem: please  remove the buil.DEBUG;
```
if  len(changeEntry.AppliedBlocks) == 0 && len(changeEntry.RevertedBlocks) != 0 {

File: modules/consensus/consensusdb.go [hunk 7]
```diff
@@ -317,53 +316,53 @@ func addBlockStakeOutput(tx *bolt.Tx, id types.BlockStakeOutputID, sfo types.Blo
 	// Sanity check - should not be adding a blockstake output with a value of
 	// zero.
 	if build.DEBUG && sfo.Value.IsZero() {
```
robvanmieghem: remove the build.DEBUG

modules/consensus/consensusdb.go
```diff
@@ -317,53 +316,53 @@ func addBlockStakeOutput(tx *bolt.Tx, id types.BlockStakeOutputID, sfo types.Blo
...
-		panic("zero value blockstake being added")
+		build.Severe("zero value blockstake being added")
 	}
 	// Sanity check - should not be adding an item already in the db.
 	if build.DEBUG && blockstakeOutputs.Get(id[:]) != nil {
```
robvanmieghem: remove the build.DEBUG

modules/consensus/consensusdb.go
```diff
@@ -317,53 +316,53 @@ func addBlockStakeOutput(tx *bolt.Tx, id types.BlockStakeOutputID, sfo types.Blo
...
-		panic("repeat blockstake output")
+		build.Severe("repeat blockstake output")
 	}
 	err := blockstakeOutputs.Put(id[:], siabin.Marshal(sfo))
 	if build.DEBUG && err != nil {
```
robvanmieghem: remove the build.DEBUG

cmd/rivinec/main.go
robvanmieghem: Ineed but the intention is different as the comment indicates, it also asks the user to file a bug.

modules/consensus/diffs.go
```diff
@@ -28,22 +27,22 @@ func commitDiffSetSanity(tx *bolt.Tx, pb *processedBlock, dir modules.DiffDirect
 
 	// Diffs should have already been generated for this node.
 	if !pb.DiffsGenerated {
-		panic(errDiffsNotGenerated)
+		build.Severe(errDiffsNotGenerated)
```
robvanmieghem: I would use Critical here

modules/consensus/diffs.go
```diff
@@ -28,22 +27,22 @@ func commitDiffSetSanity(tx *bolt.Tx, pb *processedBlock, dir modules.DiffDirect
...
 	if dir == modules.DiffApply {
 		parent, err := getBlockMap(tx, pb.Block.ParentID)
-		if build.DEBUG && err != nil {
-			panic(err)
+		if err != nil {
+			build.Severe(err)
 		}
 		if parent.Block.ID() != currentBlockID(tx) {
-			panic(errWrongAppliedDiffSet)
+			build.Severe(errWrongAppliedDiffSet)
```
robvanmieghem: I would use Critical here

modules/consensus/diffs.go
```diff
@@ -28,22 +27,22 @@ func commitDiffSetSanity(tx *bolt.Tx, pb *processedBlock, dir modules.DiffDirect
...
 		}
 	} else {
 		if pb.Block.ID() != currentBlockID(tx) {
-			panic(errWrongRevertDiffSet)
+			build.Severe(errWrongRevertDiffSet)
```
robvanmieghem: I would use Critical here

modules/consensus/fork_helpers_test.go
```diff
@@ -32,7 +33,7 @@ func (cs *ConsensusSet) dbForkBlockchain(pb *processedBlock) (revertedBlocks, ap
 		return nil
 	})
 	if updateErr != nil {
-		panic(updateErr)
+		build.Severe(updateErr)
```
robvanmieghem: I would use Critical here

modules/consensus/outputs.go
```diff
@@ -12,7 +13,7 @@ func (cs *ConsensusSet) GetCoinOutput(id types.CoinOutputID) (co types.CoinOutpu
 		return nil
 	})
 	if dbErr != nil {
-		panic(dbErr)
+		build.Severe(dbErr)
```
robvanmieghem: I would use Critical here

robvanmieghem:
Since not only the statenents where if build.Debug were refactored, I would remove the if build.Debug from the Crtical funton to keep the behaviour consistent.

crypto/signatures_test.go
DylanVerstraete: Errors were a bit longer since `Severe error:` .. Gets added to the error string. This made that the error string was cut of in debug mode. 

Example of such cut off string:
`could not decode type crypto.Signature: Severe error unexpected EO +5more`

Since we can't match on this string we match using a contain substring function.

DylanVerstraete:
As @LeeSmet said, removing all the check with `build.Debug && anotherExpensiveOperation`
will reduce performance since this `anotherExpensiveOperation` will always be checked even if the build isn't `debug`. Any thoughts @robvanmieghem ?

robvanmieghem:
This is true , question is if `anotherExpensiveOperation` is really that expensive, which requires  a thorough knowledge of the code . I prefer a sanity check that panics instead of an unseen error.

LeeSmet:
It's not really about a single check, but now we will always do all debug checks even in production. Maybe we should test this by letting a production build sync and time it, and letting a debug build sync and time it, to get an idea of how much of a performance penalty this is going to be (if any).

pkg/encoding/siabin/marshal_test.go
```diff
@@ -130,7 +133,9 @@ func TestDecode(t *testing.T) {
 
 	// bad boolean
 	err := Unmarshal([]byte{3}, new(bool))
-	if err == nil || err.Error() != "could not decode type bool: boolean value was not 0 or 1" {
+	contains = strings.Contains(err.Error(), "could not decode type bool: Severe")
```
GlenDC: Put it back next to `err == nil`, currently this will panic when `err == nil`, as you're trying to call a method on a `nil` object.

pkg/encoding/siabin/marshal_test.go
```diff
@@ -142,26 +147,34 @@ func TestDecode(t *testing.T) {
 
 	// unknown type
 	err = Unmarshal([]byte{1, 2, 3}, new(map[int]int))
-	if err == nil || err.Error() != "could not decode type map[int]int: unknown type" {
+	contains = strings.Contains(err.Error(), "could not decode type map[int]int:")
```
GlenDC: Put it back next to `err == nil`, currently this will panic when `err == nil`, as you're trying to call a method on a `nil` object.

pkg/encoding/siabin/marshal_test.go
```diff
@@ -142,26 +147,34 @@ func TestDecode(t *testing.T) {
...
+
+	if err == nil || !contains {
 		t.Error("expected unknown type error, got", err)
 	}
 
 	// big slice (larger than MaxSliceSize)
 	err = Unmarshal(EncUint64(MaxSliceSize+1), new([]byte))
-	if err == nil || err.Error() != "could not decode type []uint8: encoded slice is too large" {
+	contains = strings.Contains(err.Error(), "could not decode type []uint8: Severe error")
```
GlenDC: Put it back next to `err == nil`, currently this will panic when `err == nil`, as you're trying to call a method on a `nil` object.

pkg/encoding/siabin/marshal_test.go
```diff
@@ -142,26 +147,34 @@ func TestDecode(t *testing.T) {
...
+
+	if err == nil || !contains {
 		t.Error("expected large slice error, got", err)
 	}
 
 	// massive slice (larger than MaxInt32)
 	err = Unmarshal(EncUint64(1<<32), new([]byte))
-	if err == nil || err.Error() != "could not decode type []uint8: encoded slice is too large" {
+	contains = strings.Contains(err.Error(), "could not decode type []uint8: Severe error")
```
GlenDC: Put it back next to `err == nil`, currently this will panic when `err == nil`, as you're trying to call a method on a `nil` object.

pkg/encoding/siabin/marshal_test.go
```diff
@@ -142,26 +147,34 @@ func TestDecode(t *testing.T) {
...
+
+	if err == nil || !contains {
 		t.Error("expected large slice error, got", err)
 	}
 
 	// many small slices (total larger than maxDecodeLen)
 	bigSlice := strings.Split(strings.Repeat("0123456789abcdefghijklmnopqrstuvwxyz", (MaxSliceSize/16)-1), "0")
 	err = Unmarshal(Marshal(bigSlice), new([]string))
-	if err == nil || err.Error() != "could not decode type []string: encoded object exceeds size limit" {
+	contains = strings.Contains(err.Error(), "could not decode type []string:")
```
GlenDC: Put it back next to `err == nil`, currently this will panic when `err == nil`, as you're trying to call a method on a `nil` object.

pkg/encoding/siabin/marshal_test.go
```diff
@@ -175,7 +188,9 @@ func TestDecode(t *testing.T) {
 	}
 	// special case, not covered by testStructs
 	err = dec.Decode(new([3]byte))
-	if err == nil || err.Error() != "could not decode type [3]uint8: EOF" {
+	contains = strings.Contains(err.Error(), "could not decode type [3]uint8")
```
GlenDC: Put it back next to `err == nil`, currently this will panic when `err == nil`, as you're trying to call a method on a `nil` object.

GlenDC:
> It's not really about a single check, but now we will always do all debug checks even in production. 

- OK, it does mean that you now indeed will have those if checks even when not in DEBUG, but as Rob says, I do think having some logging even in production for those situations which we deem severe anyhow, is not a luxury;
- `build.DEBUG` will still be removed from compilation whether it is in a function or not.
- it is true we will now do the other checks, but most of them just check something cheap enough to compute, with the exception of a few `boltdb` bucket calls, but even those are cheap enough that I don't expect the performance to go suddenly down;

I would just merge it as-is, and if you do start to see performance drops, we can measure and if needed define a function such as:

`func SevereIf(f func() bool, v ...interface{})`

PR #507, opened 2019-03-18
Fix #504: Require miner fees for regular txns

### LINKED ISSUE #504: is it OK that miner fees are currently opt-out in v1 transactions
We do validate that when miner fees are defined, that they are equal to or higher than the minimum value.  We also check for a block if the sum of all fees are as expected. However the expected sum really just takes into account the fees that are defined, meaning if you don't define any fee, it will just happily accept this.

In order to reproduce this, try to create a custom transaction using a script or CLI, sign it, and send it, you'll see it will accept a transaction without having to sign it. (Note that when using CLI you will have to modify the created transaction a bit, as our code for creating a coin transaction does validate, but that is not the official validation, so as long as you avoid that, you can happily sign and push a transaction without paid miner fees into the blockchain.

That this bug (feature?:) has remained undetected for so long is surprising to say the least. Until now this hasn't been an issue as everybody is using one of our clients (JSX/CLI/Mobile) which all define transactions like good boys.

Now that this bug (feature?:) is detected however, one might actually wonder if this "opt-out" experience is really so bad. Normally these fees are to help incentivize the block creators to:

1) include the transaction in a block
2) if too many transactions are available to include, take the highest paid ones

In our case however the transaction fees go for **100%** to the TF Foundation, and thus I wonder if it is such a bad idea to keep it opt-out.

Either way, this became an open question out of the blue that we should fix ASAP.

LeeSmet:
Require that all transactions define a high enough fee (except for transactions explicitly created by block creators for the POBS protocol). Previously it was possible to create transactions without fees (though 0 value fees were correctly rejected)

### DIFF
--- a/modules/consensus/diffs.go
+++ b/modules/consensus/diffs.go
@@ -156,12 +156,12 @@ func (cs *ConsensusSet) generateAndApplyDiff(tx *bolt.Tx, pb *processedBlock) er
 	// Validate and apply each transaction in the block. They cannot be
 	// validated all at once because some transactions may not be valid until
 	// previous transactions have been applied.
-	for _, txn := range pb.Block.Transactions {
+	for idx, txn := range pb.Block.Transactions {
 		err := validTransaction(tx, txn, types.TransactionValidationConstants{
 			BlockSizeLimit:         cs.chainCts.BlockSizeLimit,
 			ArbitraryDataSizeLimit: cs.chainCts.ArbitraryDataSizeLimit,
 			MinimumMinerFee:        cs.chainCts.MinimumTransactionFee,
-		}, pb.Height, pb.Block.Timestamp)
+		}, pb.Height, pb.Block.Timestamp, cs.isBlockCreatingTx(idx, pb.Block))
 		if err != nil {
 			cs.log.Printf("WARN: block %v cannot be applied: tx %v is invalid: %v",
 				pb.Block.ID(), txn.ID(), err)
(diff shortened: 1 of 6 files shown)


modules/consensus/diffs.go
```diff
@@ -200,3 +200,46 @@ func (cs *ConsensusSet) generateAndApplyDiff(tx *bolt.Tx, pb *processedBlock) er
 
 	return blockMap.Put(bid[:], siabin.Marshal(*pb))
 }
+
+// isBlockCreatingTx checks if a transaction at a given index in the block is considered to
```
GlenDC: Looks good, but why do we need to define such a function. Does the regular POBS than not use such a function to check if the Tx is indeed a block creation Tx??

types/validtransaction.go
```diff
@@ -71,6 +72,11 @@ func TransactionFollowsMinimumValues(t Transaction, minimumMinerFee Currency) er
 			return ErrTooSmallMinerFee
 		}
 	}
+	// also reject abscent miner fees is the minimum fee is non zero
+	// block creation transactions are exempt from this rule
+	if len(t.MinerFees) == 0 && !IsBlockCreatingTx && !minimumMinerFee.IsZero() {
```
GlenDC: Personally I would put `!IsBlockCreatingTx` prior to the len check

modules/consensus/diffs.go
LeeSmet: No, currently the consensus only checks if the the blockstake output indicated in the header is respent in the block, though it does not care in what context that happens. I.E. it is legal to use a transfer of a blockstake to someone else in a regular transaction (which does require miner fees_ as the spending of the blockstake for the POBS algorithm.

GlenDC: Isn’t it weird that there we are not just as strict?

LeeSmet: No because the POBS algorithm only cares that ANY valid blockstake output is respent. However here, we need to verify that the transaction we use is created by the blockcreator to prove that he owns the bockstakes in the current version of the protocol, as then the block creator does not need to pay a minerfee. As an example, I might be selling my blockstake(s) to you, which I do in a regular transaction. So I create a blockstake input and output, and also a coin input (and possibly output) to pay the minerfee, and push this tx to the txpool. My block creator can then still use these blockstakes to create the next block (since I still own them untill said block is created). This is however, a regular transaction, and not a special transaction used for the sole purpose of creating the block. The POBS algorithm does not care about that, but we do care here since only transactions created to prove ownership op blockstakes don't require a minerfee

PR #508, opened 2019-03-20
Added support for bootstrap-peers flag in cli

DylanVerstraete:
This links to https://github.com/threefoldtech/rivine/issues/502.
Adds support for bootstrap-peers flag in Rivine daemon CLI. 

@robvanmieghem is there any way to test this flag in Rivine?

### DIFF
--- a/cmd/rivined/commands.go
+++ b/cmd/rivined/commands.go
@@ -15,7 +15,7 @@ import (
 )
 
 type commands struct {
-	cfg           daemon.Config
+	cfg           ExtendedDaemonConfig
 	moduleSetFlag daemon.ModuleSetFlag
 }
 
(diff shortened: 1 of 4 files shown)


GlenDC:
You can test it by starting a block creator node on your own devnet with ‘—no-bootstrap’, let it create some blocks. Than start a 2nd node on the same devnet without specifying the ‘—no-boostrap’ flag but do specify the first nodes address as the value for the ‘bootstrap-peers flag.

Check with @leesmet for questions.

DylanVerstraete:
Tested these changes and the node synced up with the bootstrap peer.

GlenDC:
@DylanVerstraete you misunderstood the goal of this issue. What you did has not much value as it just adds some custom thing to the rivine daemon application, but that was not the point. The goal is that you are able to override the bootstrap peers in a rivine-based blockchain by default, without having to specifying such logic yourself.

Closing this PR as this is going nowhere, we'll look at it together tomorrow.

PR #511, opened 2019-03-25
add certmagic package to host explorer frontend with https

### LINKED ISSUE #505: Host the explorer html ourselves
Drop the need for caddy, investigate if we can import the caddy code for let's encrypt license wise

DylanVerstraete:
fixes #505 

Example of how to integrate in Tfchain: 

```
// Start serving frontend if there is such a folder
go func() {
	domains := []string{""}
	err = explorer.ServeFrontend(true, domains, "")
	if err != nil {
		servErrs <- err
	}
}()
```


robvanmieghem:
I like the certmagic integration but not the hardcoding of LetsEncryptStagingCA  Ca and the port, it also uses a different domain/port now than the  explorer rest api

DylanVerstraete:
Added flags to `rivined`:

* `c` "stagingCertificate", indicates if a staging certificate should be used.
* `e` "hostExplorer", indicates if we should host the explorer ourself.
* `caddy-domains` list of caddy domains for our certificate.

It also now serves the frontend and backend on the same host/port.

Needs to be tested in production right now.

GlenDC:
@LeeSmet please give a proper review of this PR ASAP.

PR #512, opened 2019-03-27
make minting extension more compatible for tfchain

DylanVerstraete:
On master branch in Rivine the minting feature is unusable because its creates a new TransactionDB. This will conflict with the TransactionDB that will be created in TFchain. This pull request extends the usability of the TransactionDB by opening the existing database and adding the `minting buckets` to a TransactionDB instead of creating a seperate TransactionDB. 

The readme explains how you can integrate this feature in Rivine or Goldchain for example.


DylanVerstraete:
Closing because this implementation will not work

PR #513, opened 2019-04-01
Minting extensions that plugs in to the consensus

DylanVerstraete:
This PR proposes a new architecture. A plugin system that can be subscribed to the consensus. A readme explains how this is implemented.


extensions/minting/client/walletcmd.go
```diff
@@ -0,0 +1,213 @@
...
+	err = condition.UnmarshalJSON([]byte(str))
+	if err != nil {
+		return types.UnlockConditionProxy{}, fmt.Errorf(
+			"condition has to be UnlockHash or JSON-encoded UnlockCondition, output %q is neither", str)
+	}
+	return
+}
+
+// RandomTransactionNonce creates a random Transaction nonce
+func RandomTransactionNonce() (nonce minting.TransactionNonce) {
```
GlenDC: I wouldn't put the `TransactionNonce` under `minting` though, as it is not specific to `minting`.

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+	"fmt"
+
+	bolt "github.com/rivine/bbolt"
+	persist "github.com/threefoldtech/rivine/persist"
+	"github.com/threefoldtech/rivine/pkg/encoding/siabin"
+	"github.com/threefoldtech/rivine/types"
+)
+
+const (
+	pluginDbVersion = "1.0.0.0"
```
GlenDC:
```suggestion
	pluginDBVersion = "1.0.0.0"
```

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+	pluginDbHeader  = "mintingPlugin"
```
GlenDC:
```suggestion
	pluginDBHeader  = "mintingPlugin"
```

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+		err := bucket.Put(EncodeBlockheight(0), mintcond)
+		if err != nil {
+			return persist.Metadata{}, fmt.Errorf("failed to store genesis mint condition: %v", err)
+		}
+		metadata = &persist.Metadata{
+			Version: pluginDbVersion,
+			Header:  pluginDbHeader,
+		}
+	} else if metadata.Version != pluginDbVersion {
+		// TODO upgrade and set metadata to new version
```
GlenDC: why the TODO?

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+		genesisMintCondition: genesisMintCondition,
+	}
+	types.RegisterTransactionVersion(TransactionVersionMinterDefinition, MinterDefinitionTransactionController{MintConditionGetter: p})
+	types.RegisterTransactionVersion(TransactionVersionCoinCreation, CoinCreationTransactionController{MintConditionGetter: p})
+	return p
+}
+
+// InitBucket initializes the Bucket for the first time
+func (p *Plugin) InitBucket(metadata *persist.Metadata, name string, bucket *bolt.Bucket, db *persist.BoltDatabase) (persist.Metadata, error) {
+	p.database = db
```
GlenDC: This is a defeating the point entirely, if it has access to the BoltDB, then that means it has to know the structure of how its data is organised, which defeats the entire point of this abstraction (e.g. giving the bucket). I do understand why you are giving the DB as you otherwise have no way to get the data, however there must be a cleaner approach. Perhaps instead of giving a bucket, name (btw you look for the bucket in the name, but isn't that plugin-bucket, a child bucket of another bucket (the `plugins` bucket?)).

Perhaps we can give a plugin a storage interface, with an API such as:

```go
type PluginStorage interface {
     View(bucket *bolt.Bucket) error
     Update(bucket *bolt.Bucket) error
}
```

So that the consensus takes care of all this, and again we managed to abstract away. In order to allow the plugin to store this storage value, it does need to be ensured that when we share it, underlying, the bolt database instance, that it is safe to do so (e.g. that we do not release it until all users released its ownership of it).

GlenDC: But certainly don't share this raw instance just like that.

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+	} else if metadata.Version != pluginDbVersion {
+		// TODO upgrade and set metadata to new version
+		return persist.Metadata{}, errors.New("There is only 1 version of this plugin, version mismatch")
+	}
+	return *metadata, nil
+}
+
+// ApplyBlock applies a block's minting transactions to the minting bucket.
+func (p *Plugin) ApplyBlock(block types.Block, height types.BlockHeight, _ *persist.LazyBoltBucket) error {
+	err := p.database.Update(func(tx *bolt.Tx) (err error) {
```
GlenDC: point was to use the bucket, otherwise what is the point. Now you open another transaction which I'm not sure is even the sanest thing to do. If you go however for the storage approach than you can drop the bucket from the signature (and even that entire type).

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// RevertBlock reverts a block's minting transaction from the minting bucket
+func (p *Plugin) RevertBlock(block types.Block, height types.BlockHeight, bucket *persist.LazyBoltBucket) error {
+	err := p.database.View(func(tx *bolt.Tx) (err error) {
```
GlenDC: point was to use the bucket, otherwise what is the point. Now you open another transaction which I'm not sure is even the sanest thing to do. If you go however for the storage approach than you can drop the bucket from the signature (and even that entire type).

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+		// TODO upgrade and set metadata to new version
+		return persist.Metadata{}, errors.New("There is only 1 version of this plugin, version mismatch")
+	}
+	return *metadata, nil
+}
+
+// ApplyBlock applies a block's minting transactions to the minting bucket.
+func (p *Plugin) ApplyBlock(block types.Block, height types.BlockHeight, _ *persist.LazyBoltBucket) error {
+	err := p.database.Update(func(tx *bolt.Tx) (err error) {
+		bucket := tx.Bucket([]byte(p.name))
```
GlenDC: I thought the plugin bucket is a child bucket of the `/plugins` bucket instead of being a root bucket?

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// RevertBlock reverts a block's minting transaction from the minting bucket
+func (p *Plugin) RevertBlock(block types.Block, height types.BlockHeight, bucket *persist.LazyBoltBucket) error {
+	err := p.database.View(func(tx *bolt.Tx) (err error) {
+		bucket := tx.Bucket([]byte(p.name))
```
GlenDC: I thought the plugin bucket is a child bucket of the `/plugins` bucket instead of being a root bucket?

extensions/minting/minting.go
```diff
@@ -0,0 +1,200 @@
...
+		return err
+	}
+	return nil
+}
+
+// GetActiveMintCondition implements types.MintConditionGetter.GetActiveMintCondition
+func (p *Plugin) GetActiveMintCondition() (types.UnlockConditionProxy, error) {
+	var mintCondition types.UnlockConditionProxy
+	err := p.database.View(func(tx *bolt.Tx) (err error) {
+		bucket := tx.Bucket([]byte(p.name))
```
GlenDC: I thought the plugin bucket is a child bucket of the `/plugins` bucket instead of being a root bucket?

persist/boltdb.go
```diff
@@ -100,3 +101,140 @@ func OpenDatabase(md Metadata, filename string) (*BoltDatabase, error) {
 
 	return boltDB, nil
 }
+
+// LazyBoltBucket is a lazy implementation of the bolt DB,
+// allowing you to only actually get the bucket from bolt db if you need it.
+type LazyBoltBucket struct {
```
GlenDC: This can be dropped if we go for the storage interface approach.

extensions/minting/minting.go
GlenDC: If you have a better idea than the `PluginStorage` also cool though, but this thing is already a hack, but having the requirement of the _Plugins_ to know how the data is stored defeats its entire point.

DylanVerstraete: @GlenDC FYI. From a bucket you can reference to a `TX` which can reference to the underlying database.

GlenDC: Yes obviously we cannot stop them (easily) from doing something stupid. As we are the only users that is OK. Most important (here) is that the most obvious paths are also the correct paths.

persist/boltdb.go
DylanVerstraete: Why can we drop this ?

extensions/minting/client/walletcmd.go
DylanVerstraete: Where would we put it then?

modules/consensus/accept.go
```diff
@@ -176,6 +177,35 @@ func (cs *ConsensusSet) addBlockToTree(b types.Block) (ce changeEntry, err error
...
 			return err
 		}
+
+		// update consensus change id for all plugins
+		if len(cs.plugins) > 0 {
+			consensusChangeID := ce.ID()
+			for name := range cs.plugins {
+				metadataBucket := tx.Bucket(bucketPluginsMetadata)
+				if metadataBucket == nil {
+					return ErrPluginGhostMetadata
```
GlenDC: Wrong error, should be an error saying the root plugins metadata folder is missing, a bug.

By the way (!!!!!!!) This Metadata bucket should be a child bucket of the root `Plugins` bucket. That is not the case here. Make sure it is. And not only here, but in all other places where you get/create the `Metadata` bucket!!!

modules/consensus.go
```diff
@@ -242,6 +269,8 @@ type (
 
 		// GetBlockStakeOutput takes a blockstake output ID and returns the appropriate blockstake output
 		GetBlockStakeOutput(types.BlockStakeOutputID) (types.BlockStakeOutput, error)
+
+		RegisterPlugin(name string, plugin ConsensusSetPlugin, cancel <-chan struct{}) error
```
GlenDC: Why you did not add the other function as well (`UnregisterPlugin`)??

extensions/minting/readme.md
```diff
@@ -0,0 +1,47 @@
...
+## Usage
+
+### Daemon
+
+After creating the consensus module and registering the http handlers you can use following snippet
+
+```golang
+condition := AnyUnlockCondition
+plugin := minting.NewMintingPlugin(types.UnlockConditionProxy{Condition: condition})
+    err = cs.RegisterPlugin("minting", plugin, cancel)
```
GlenDC: wrong indent level

extensions/minting/readme.md
```diff
@@ -0,0 +1,47 @@
+# Rivine minting extensions
+
+## Usage
+
+### Daemon
+
+After creating the consensus module and registering the http handlers you can use following snippet
+
+```golang
+condition := AnyUnlockCondition
```
GlenDC: Give a proper UnlockCondtion as example (with comments to define what other conditions can be used as well)

extensions/minting/minting.go
```diff
@@ -0,0 +1,172 @@
...
+			continue // ignore most common Tx
+		}
+		// check the version and handle the ones we care about
+		switch rtx.Version {
+		case TransactionVersionMinterDefinition:
+			mdtx, err := MinterDefinitionTransactionFromTransaction(*rtx)
+			if err != nil {
+				return fmt.Errorf("unexpected error while unpacking the minter def. tx type: %v" + err.Error())
+			}
+			err = bucket.Put(EncodeBlockheight(height), siabin.Marshal(mdtx.MintCondition))
```
GlenDC: Given you start from a clean slate here, I would make a child bucket with mint conditions and put them all in there. This way, it would allow you in the future to add some other data besides the height-to-condition mapping that we have now. Otherwise you can no longer do that, as our searching trick would no longer work.

(!!) Make sure that this is applied correctly at all places of this file, otherwise you'll introduce bugs.

modules/consensus/plugin_storage.go
```diff
@@ -0,0 +1,39 @@
...
+	"errors"
+
+	"github.com/threefoldtech/rivine/persist"
+
+	bolt "github.com/rivine/bbolt"
+)
+
+// A PluginViewStorage struct is a definition for the storage tool for a plugin bucket
+type PluginViewStorage struct {
+	db   *persist.BoltDatabase
```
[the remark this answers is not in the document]
DylanVerstraete: Whats do you specifically have in mind for this?

extensions/minting/client/walletcmd.go
GlenDC: Somewhere in the `types` package, so that it can be used by anyone. You can do this in a follow-up PR.

PR #525, opened 2019-05-10
Adds retry loop for connecting to bootstrap nodes

### LINKED ISSUE #517: Periodically try to connect to bootstrap nodes
Right now the gateway will try to automatically connect to nodes it previously was connected to untill it has a proper amount of outbound peers. However in case a node does not have internet connectivity when the daemon is started, it will be unable to resolve the bootstrap nodes, won't connect to them, and as a result have an empty list of known nodes. This causes the daemon to be peerless permanently, unless a node is manually connected.

It would be better if next to connecting to previously known nodes, the gateway also attempts to connect to the bootstrap nodes if it does not have a sufficient amount of known nodes

DylanVerstraete:
fixes #517

We wait 1 minute before trying to connect again to a bootstrap node which we couldn't connect to.

### DIFF
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -347,7 +347,7 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
 
 func (g *Gateway) connectToBootstapPeers(closeChan chan struct{}, bootstrapPeers []modules.NetAddress) {
 	defer close(closeChan)
-
+g.startConnectingToBootstrapPeers(bootstrapPeers)
 	for {
 		select {
 		// If gateway stop, close the closeChannel
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -274,6 +274,11 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
 			if err := addr.TryNameResolution(); err != nil {
 				// Bootstrap nodes can still be in IP:PORT notation so we might still be able to continue
 				g.log.Debugf("Bootstrap node [%v] address resolution failed: %v", addr, err)
+
+				// Start a retry loop for every failed bootstrap node connection
+				go func() {
+					g.connectToBootstapPeer(addr)
+				}()
 			}
 			err := g.addNode(addr)
 			if err != nil && err != errNodeExists {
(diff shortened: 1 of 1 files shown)


modules/gateway/gateway.go
```diff
@@ -343,5 +348,25 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
 	return g, nil
 }
 
+func (g *Gateway) connectToBootstapPeer(bootstrapPeer modules.NetAddress) error {
+	// Wait 1 minute
+	time.Sleep(5 * time.Minute)
```
LeeSmet: In general, 1 minute != 5 * time.Minute

modules/gateway/gateway.go
```diff
@@ -343,5 +348,25 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
...
+	g.log.Debugf("retrying connection to bootstrap node [%v]", bootstrapPeer)
+
+	if err := bootstrapPeer.TryNameResolution(); err != nil {
+		// Bootstrap nodes can still be in IP:PORT notation so we might still be able to continue
+		g.log.Debugf("Bootstrap node [%v] address resolution failed: %v", bootstrapPeer, err)
+
+		// Recursive retry loop
+		g.connectToBootstapPeer(bootstrapPeer)
```
LeeSmet: There is a couple of issues here:

1. The error returned is not caught and handled
2. There is no return, which means that eventually when this function call returns, this function will progress, even though that is likely not intended
3. Since this function recurses indefinitely, it could potentially bring down a healthy daemon via a stack overflow, if a single bootstrap node is down or otherwise unreachable

modules/gateway/gateway.go
```diff
@@ -343,5 +348,25 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
 	return g, nil
 }
 
+func (g *Gateway) connectToBootstapPeer(bootstrapPeer modules.NetAddress) error {
```
LeeSmet: Please create a closeChan and pass it as argument. Then, instead of the loop plus sleep, change that to an infinite loop, with a first block which selects over the timeout, and the close chan. In case the closechan receives a value, return. In case the timeout fires, perform the logic for adding peers. In the new function, there are already examples of creating such a close chan, and attaching it to the threadpool used by the gateway. This allows the goroutine to be cleaned up should the gateway exit (but not the daemon), otherwise we will have this lingering goroutine throwing errors

modules/gateway/gateway.go
```diff
@@ -274,6 +274,11 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
 			if err := addr.TryNameResolution(); err != nil {
 				// Bootstrap nodes can still be in IP:PORT notation so we might still be able to continue
 				g.log.Debugf("Bootstrap node [%v] address resolution failed: %v", addr, err)
+
+				// Start a retry loop for every failed bootstrap node connection
+				go func() {
+					g.connectToBootstapPeer(addr)
```
LeeSmet: This spawns a goroutine per bootstrap peer, which is generally a bit wastefull. We can do this synchronously in a single goroutine for every peer. 

Error is not caught

modules/gateway/gateway.go
DylanVerstraete: Oops that's a typo

LeeSmet:
It's a little annoying that this only connects after a minute, other than that looks fine for me

modules/gateway/gateway.go
```diff
@@ -343,5 +345,36 @@ func New(addr string, bootstrap bool, persistDir string, bcInfo types.Blockchain
 	return g, nil
 }
 
+func (g *Gateway) connectToBootstapPeers(closeChan chan struct{}, bootstrapPeers []modules.NetAddress) {
+	defer close(closeChan)
+
```
LeeSmet:
```suggestion
g.startConnectingToBootstrapPeers(bootstrapPeers)
```

Add method call before loop so we immediately start to connect

PR #565, opened 2019-07-29
Improve RPC rate limiting

### LINKED ISSUE #519: Transactionpool propagation slow
When pushing a large amount of transactions to the pool, it takes a long time to propagate them to other nodes (about 1K small (less than 100 bytes ) transactions per hour).

### LINKED ISSUE #564: Rework gateway RPC handling
Right now the gateway uses a sleeping mechanism to ensure that a peer can't flood it with requests. After every accepted RPC call, the gateway waits for a predefined amount of time (defined as a build time constant). The constant in question is defined here: https://github.com/threefoldtech/rivine/blob/master/modules/gateway/consts.go#L81. For all intents and purposes, it is set to 3 seconds in any deployed network. The usage of this const is seen here: https://github.com/threefoldtech/rivine/blob/master/modules/gateway/rpc.go#L181. The idea here seems simple: to prevent a buggy or malicious peer from launching a denial of service attack against our node, we only perform an rpc call every x amount of time. This prevents remote nodes from exhausting the CPU or memory.

The problem is that waiting 3 seconds for every RPC call is in itself a DOS vector, especially once there is a somewhat large amount of traffic on the network. After all, it means that if multiple transactions are posted with less than 3 seconds between them for a longer period of time, the node might built up a backlog. This is somewhat mitigated by the fact that the delay is per peer, so if there is a flood of valid transactions, these can still somewhat propagate naturally as not every peer will have the same txes available at the same time. Regardless, it is still less than optimal.

The issue is worsened even more by the fact that ALL node communication happens through RPC calls. More specifically, created blocks are also propagated like this. This means that, in periods where there is a lot of transaction traffic, the chances of forks in the network increases significantly. While this is in itself also not a problem, it is a situation we might want to avoid if at all possible. The problem is: assume 100 transactions are published at once. It will take 300 seconds by default (5 mins) to propegate all those txes to our block creator from the node which received the txes. Now if the block time is less than 5 minutes, there is likely already a block created somewhere in the network. But since all our peers are still sending transactions, the block does not reach our block creator until all these transactions (actually their RPC calls) have been processed. As a result, we will be attempting to create blocks on a what is no longer the tip of the chain, causing a fork if we do find one.

To resolve this, I propose some kind of token bucket rate limiting, more specifically the following:

Once a peer is created in the gateway, we give it a buffered channel (buffer size can be configureable but for now lets leave it at one). Just before we accept the rpc call, we perform a read operation on the channel. This operations blocks the accept unless a token is present in the channel. After that, the RPC call is changed to defer writing a token back into the channel (i.e. the token is recycled when done). This way, we can control the amount of concurrent RPC calls (size of the buffer), while allowing new calls to be accepted as soon as we are idle again.

Since we still block the main accept loop, ordering should be preserved (as we block on accepting the next RPC call until the previous one is done), though this might need to be examined more in depth in case multiple RPC calls are allowed at once.

This would solve #519

LeeSmet:
Replace the sleep rate limiting implementation with a token bucket version. See #564 for details

Fix #519 
Close #564

### DIFF
--- a/modules/gateway/peers.go
+++ b/modules/gateway/peers.go
@@ -37,6 +37,8 @@ func (s insufficientVersionError) Error() string {
 type peer struct {
 	modules.Peer
 	sess streamSession
+	// rate limiting channel
+	token chan struct{}
 }
 
 // sessionHeader is sent as the initial exchange between peers.
@@ -179,7 +181,12 @@ func (g *Gateway) managedAcceptConnPeer(conn net.Conn, remoteInfo remoteInfo) er
 			NetAddress: remoteAddr,
 			Version:    remoteInfo.Version,
 		},
-		sess: newSmuxServer(conn),
+		sess:  newSmuxServer(conn),
+		token: make(chan struct{}, g.concurrentRPCPerPeer),
+	}
+	for i := 0; uint64(i) < g.concurrentRPCPerPeer; i++ {
+		// Fill the channel wit htokens
+		peer.token <- struct{}{}
 	}
 
 	g.mu.Lock()
@@ -661,15 +668,22 @@ func (g *Gateway) managedConnect(addr modules.NetAddress) error {
 	g.mu.Lock()
 	defer g.mu.Unlock()
 
-	g.addPeer(&peer{
+	peer := &peer{
 		Peer: modules.Peer{
 			Inbound:    false,
 			Local:      addr.IsLocal(),
 			NetAddress: addr,
 			Version:    remoteInfo.Version,
 		},
-		sess: newSmuxClient(conn),
-	})
+		sess:  newSmuxClient(conn),
+		token: make(chan struct{}, g.concurrentRPCPerPeer),
+	}
+	for i := 0; uint64(i) < g.concurrentRPCPerPeer; i++ {
+		// Fill the channel with tokens
+		peer.token <- struct{}{}
+	}
+
+	g.addPeer(peer)
 	g.addNode(addr)
 	g.nodes[addr].WasOutboundPeer = true
 
--- a/cmd/rivined/daemon.go
+++ b/cmd/rivined/daemon.go
@@ -22,6 +22,11 @@ import (
 	"github.com/threefoldtech/rivine/pkg/daemon"
 )
 
+const (
+	// maxConcurrentRPC is the maximum number of concurrent RPC calls allowed for a single daemon
+	maxConcurrentRPC = 1
+)
+
 func runDaemon(cfg daemon.Config, networkCfg daemon.NetworkConfig, moduleIdentifiers daemon.ModuleIdentifierSet) error {
 	// Print a startup message.
 	fmt.Println("Loading...")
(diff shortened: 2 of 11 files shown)


modules/gateway/peers.go
```diff
@@ -661,15 +668,22 @@ func (g *Gateway) managedConnect(addr modules.NetAddress) error {
...
 			Local:      addr.IsLocal(),
 			NetAddress: addr,
 			Version:    remoteInfo.Version,
 		},
-		sess: newSmuxClient(conn),
-	})
+		sess:  newSmuxClient(conn),
+		token: make(chan struct{}, g.concurrentRPCPerPeer),
+	}
+	for i := 0; uint64(i) < g.concurrentRPCPerPeer; i++ {
```
GlenDC: Doesn't this effectively lock you by default to 1 peer per node?

LeeSmet: Not really, as the channel used is set on the peer, and later used in the accept loop on the peer. The accept loop accepts RPC calls, not new peers

LeeSmet:
The function we are interested in is this: https://github.com/threefoldtech/rivine/blob/47d7daa3ab9e07266ad8352a0ff71783bf98903a/modules/gateway/rpc.go#L127-L190

This starts function sets up the closing of the routine, before starting the main listening loop for incoming RPC's: https://github.com/threefoldtech/rivine/blob/47d7daa3ab9e07266ad8352a0ff71783bf98903a/modules/gateway/rpc.go#L165-L184

So the function does the following, in order:
 - accept a new session on the stream to the peer (effectively accept an incoming RPC call)
 - try to set the deadline to read from the session, so malicious or buggy peers can't leave a connection hanging
 - start the threadedhandlecon, which handles the actual RPC, in a new goroutine
 - sleep for the specified time (3 seconds in non-dev/testing)

The last step is the important part: after launching the RPC in its own goroutine, the main loop idles for 3 seconds. This means that:
 - RPC's taking less than 3 seconds cause idle time in the system
 - RPC's taking longer than 3 seconds cause multiple RPC's to be handled concurrently

The changes in this PR change the behavior to the following:
 - accept a new session on the stream to the peer (effectively accept an incoming RPC call)
 - try to set the deadline to read from the session, so malicious or buggy peers can't leave a connection hanging
 - take a token from the new token channel, blocking until one is available. Since the token is pulled on the main loop, we maintain the ordering from the accept on the session.
 - start the threadedhandlecon, which handles the actual RPC, in a new goroutine. Also now, add a defer which sends the token we pulled earlier back onto the token channel. Defer here since it makes sure the token is always recycled, even in case of a panic.

Operation now depends on the size of the channels buffer. In case the buffer size is 1 (which is currently the default), only 1 RPC for the peer is accepted at the same time. However, if the RPC takes less than 3 seconds and another one is waiting, there is no idling anymore, but the next call is accepted and processed. Likewise if the RPC takes more than 3 seconds, no new goroutine is started.

This means that this approach is in theory vulnerable for a DOS where an attacker sends multiple "slow" RPC's to the victim. But this was already a problem in the old implementation, as doing that would mean the victim inevitably spinned up a lot of goroutines, which then would cause memory issues. In that context, doing this with the new code will simply block other RPC's from that specific peer. Such RPC's however, can be classified as either a bug (since 3 seconds is a really long time to get anything done), or an exploitation by a malicous user. In the first case, we as developers are at fault, while in the second case, we will just seemingly ignore the offending peer. Compared to the old behavior, it seems better to just keep the malicious peer hanging than to allow him to spin up multiple goroutines on our system, consuming more cpu and memory than intended.
 -

PR #577, opened 2019-08-19
Rivinecg tool to generate blockchain code from our template repo.

DylanVerstraete:


GlenDC:
Will fix and merge this PR tomorrow

cmd/rivinecg/pkg/config/github_template.go
```diff
@@ -70,6 +71,9 @@ func writeTemplateValues(destinationDirPath string, config *Config) error {
 	fmap := template.FuncMap{
 		"formatConditionAsString": formatConditionAsString,
 	}
+	for n, f := range sprig.FuncMap() {
```
DylanVerstraete: @GlenDC What does this do?

GlenDC: sprig is a library that bundles a lot of nice functions in a library, we range here over all functions (name and functions) they provide and add them to the fmap we use on our templates

PR #592, opened 2019-09-18
bootstrap peers connect is now asynchronous

DylanVerstraete:


GlenDC:
@LeeSmet, could you please review and test this patch.

modules/gateway/gateway.go
```diff
@@ -348,34 +349,60 @@ func New(addr string, bootstrap bool, concurrentRPCPerPeer uint64, persistDir st
 }
 
 func (g *Gateway) connectToBootstapPeers(closeChan chan struct{}, bootstrapPeers []modules.NetAddress) {
-	defer close(closeChan)
+	defer closeGatewayChan(closeChan)
```
LeeSmet: What is the point of making a custom function here instead of just closing like before?

modules/gateway/gateway.go
```diff
@@ -348,34 +349,60 @@ func New(addr string, bootstrap bool, concurrentRPCPerPeer uint64, persistDir st
...
 		case <-time.After(1 * time.Minute):
-			g.startConnectingToBootstrapPeers(bootstrapPeers)
+			g.startConnectingToBootstrapPeers(closeChan, bootstrapPeers)
 		}
 	}
 }
 
-func (g *Gateway) startConnectingToBootstrapPeers(bootstrapPeers []modules.NetAddress) {
+func (g *Gateway) startConnectingToBootstrapPeers(closeChan chan struct{}, bootstrapPeers []modules.NetAddress) {
+	defer closeGatewayChan(closeChan)
```
LeeSmet: why is the channel passed here? the parent function already closes it when it is done, and this function just checks the threadgroup

modules/gateway/gateway.go
DylanVerstraete: Because both functions try to close this channel when the gateway shuts down.

modules/gateway/gateway.go
DylanVerstraete: Can you provide more information ?

LeeSmet: The current setup is as follows:

The gateway launches the `connectToBootstrapPeers` function in a goroutine and provides a channel which will inform the gateway that this goroutine has ended (once the gateway signals it is shutting down). This function keeps calling the child function `startConnectingToBootstrapPeers` to do the actual work, untill it notices that `g.threads` is closed.

When this happens, the goroutine should stop as soon as possible, and as a last step close the channel it has been passed. However if you are in the process of connecting to the bootstrap peers, you try to close the channel twice. This is not needed, and in fact bad. The child function will close the channel and return, thus providing control back to the parent function. This means that the gateway now thinks the goroutine is done, while it might still be doing work. Lets say that the parent goroutine needs to do a graceful shutdown of some resource before it ends. It might no longer have the opportunity as the gateway might have finished closing, and thus the binary might have exited.

The child function already checks in between every bootstrap peer it wants to connect to if the gateway is not shutting down. This is good. However the child function should not close the channel when this happens, as it has no control over the parent function behavior. It is up to the parent function to close the channel after it has done any cleanup it might need to do.

TL;DR: Remove the close channel call in the `startConnectionToBootstrapPeers` function. Then you have a proper control flow, in which the top level function in the goroutine is the only one which decides when it is done. This also means you can remove the channel close functions again and use the language builtin `close` call once more.

LeeSmet: Also, please rename `startConnectingToBootstrapPeers`, as the naming is confusing :)

LeeSmet: And while you are at it, `s/connectToBootstapPeers/connectToBootstrapPeers/g`

modules/gateway/gateway.go
```diff
@@ -356,23 +357,29 @@ func (g *Gateway) connectToBootstapPeers(closeChan chan struct{}, bootstrapPeers
...
 		// Start connection to bootstrapPeers after 1 minute
 		case <-time.After(1 * time.Minute):
-			g.startConnectingToBootstrapPeers(bootstrapPeers)
+			g.initialBootstrapPeersConnect(bootstrapPeers)
 		}
 	}
 }
 
-func (g *Gateway) startConnectingToBootstrapPeers(bootstrapPeers []modules.NetAddress) {
+func (g *Gateway) initialBootstrapPeersConnect(bootstrapPeers []modules.NetAddress) {
```
LeeSmet: This function is not doing any initial work, rather it simply connects to the bootstrap peers. In that regard, I suggest calling this `connectToBootstrapPeers`, and changing the parent function, which maintains a connection to bootstrap peers, to something like `ensureBootstrapPeerConnection` or something like that

GlenDC:
@DylanVerstraete, please resolve these comments such that @LeeSmet can wrap up things soon. Don't want this to be open for long.

PR #595, opened 2019-09-19
Improve stake modifier calculation

LeeSmet:
Cache the block id's and their matching parent block id's. This removes
about 2256 (in default configuration) database reads, which would read
an entire block every time, possibly loading more than a gigabyte
of data over and over again. This lazy loading removes the need to do
that, apart from the first block which will need to populate the cache.

In case a new daemon is started without any blocks, there will actually
be no database reads to calculate the stake modifier at all until the
daemon is stopped and restarted.

### DIFF
--- a/modules/consensus/consensusset.go
+++ b/modules/consensus/consensusset.go
@@ -86,6 +86,12 @@ type ConsensusSet struct {
 	// the genesis block, meaning the PoW is not very expensive.
 	dosBlocks map[types.BlockID]struct{}
 
+	// knownParentIDs keeps track of block ID's and their associated parent ID's.
+	// This allows us to find a parent block of any block, regardless whether or
+	// not it is in the active fork, without having to load all these blocks from
+	// disk and decode them.
+	knownParentIDs map[types.BlockID]types.BlockID
+
 	// checkingConsistency is a bool indicating whether or not a consistency
 	// check is in progress. The consistency check logic call itself, resulting
 	// in infinite loops. This bool prevents that while still allowing for full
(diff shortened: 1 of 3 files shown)


LeeSmet:
Developed and tested this patch in goldchain to ensure the behavior of the code remains consistent.

modules/consensus/consensusset.go
```diff
@@ -313,22 +321,86 @@ func (cs *ConsensusSet) BlockHeightOfBlock(block types.Block) (height types.Bloc
...
+			if !exists {
+				// Not found in cache, load from disk
+				// we previously updted cbID to point to the parent, so we can use it
+				// here instead of the now invalid pID
+				parent, err = getBlockMap(tx, cbID)
+				if err != nil {
+					return err
+				}
+				// save parentID for later use
+				cs.knownParentIDs[cbID] = parent.Block.Header().ParentID
```
GlenDC: assign pID first, and afterwards assign it to `cs.knownParentIDs[cbID]`

modules/consensus/proofofblockstake.go
```diff
@@ -35,20 +35,23 @@ func (cs *ConsensusSet) CalculateStakeModifier(height types.BlockHeight, block t
 	// Rollback the required amount of blocks, minus 1. This way we end up at the direct child of the
 	// block we use to calculate the stakemodifer, rather than the actual first block. Simplifies
 	// the main loop a bit
-	block, _ = cs.FindParentBlock(block, delay-1)
+	// block is not present in the DB yet so use the parent ID to count back from
```
GlenDC: rework comment please

LeeSmet:
:arrow_up: :pen:

PR #599, opened 2019-10-02
fix issue in transactionbuilder

DylanVerstraete:
closes https://github.com/threefoldtech/rivine/issues/587 & https://github.com/threefoldtech/rivine/issues/483

### DIFF
--- a/modules/wallet/money.go
+++ b/modules/wallet/money.go
@@ -270,12 +270,19 @@ func (w *Wallet) SendOutputs(coinOutputs []types.CoinOutput, blockstakeOutputs [
 
 	tpoolFee := w.chainCts.MinimumTransactionFee.Mul64(1) // TODO better fee algo
 	totalAmount := types.NewCurrency64(0).Add(tpoolFee)
+	var err error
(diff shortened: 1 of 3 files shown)


modules/wallet/money.go
```diff
@@ -270,12 +270,19 @@ func (w *Wallet) SendOutputs(coinOutputs []types.CoinOutput, blockstakeOutputs [
 
 	tpoolFee := w.chainCts.MinimumTransactionFee.Mul64(1) // TODO better fee algo
 	totalAmount := types.NewCurrency64(0).Add(tpoolFee)
+	var err error
 	txnBuilder := w.StartTransaction()
+	// Make sure to release inputs in case of an error
+	defer func() {
+		if err != nil {
+			txnBuilder.Drop()
```
GlenDC: Should this not also be done for other methods? E.g. fund?

DylanVerstraete: This is the only method that can mark outputs as locked by the wallet. There is one other method in the blockcreator but that one ignores the marking anyhow

PR #616, opened 2019-11-12
Do not continue IP discovery if the gateway is closed

robvanmieghem:
I know that checking if g.threads.IsSTopped  is easier but I prefer to use more standard go functionality instead of the next programmer having to know the custom sync code and structures.

### DIFF
--- a/modules/gateway/gateway.go
+++ b/modules/gateway/gateway.go
@@ -148,7 +148,6 @@ type Gateway struct {
 	peerTG siasync.ThreadGroup
 
 	// Utilities.
-	ctx        context.Context
 	log        *persist.Logger
 	mu         sync.RWMutex
 	persistDir string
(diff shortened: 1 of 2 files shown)


modules/gateway/gateway.go
```diff
@@ -147,6 +148,7 @@ type Gateway struct {
 	peerTG siasync.ThreadGroup
 
 	// Utilities.
+	ctx        context.Context
```
LeeSmet: I don't think the context should be added on the gateway struct. After the context is initially created, it is saved on the struct, but it is then no longer accessed. It also seems like go prefers to pass contexts as arguments, so I would just keep that approach

robvanmieghem: Thanks Lee, you are right.
My initial idea was that it could be useful later on but it indeed is just dragged along without a purpose.

PR #624, opened 2019-12-12
add test-related Makefile targets back

### LINKED ISSUE #623: travis no longer executes go tests
they were removed in 77d308cfab19b38638dc30e7cc87b1748d2e972b

GlenDC:
also removed two unit tests from the persistent pkg,
one of the two tests failed, while fixing I realized
that the code it was testing was only used in these
two tests, hence deleting it all seemed like the more
economical decision

fixes #623

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,36 @@ install-std:
 ineffassign:
 	ineffassign $(testpkgs)
 
+test:
+	go test -short -tags='debug testing' -timeout=30s $(testpkgs) -run=$(run)
+test-v:
+	go test -race -v -short -tags='debug testing' -timeout=60s $(testpkgs) -run=$(run)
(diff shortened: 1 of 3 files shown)


robvanmieghem:
Looks like the json _temp files are crewated in the persist.json.go.SaveJSon so not using the safeFile. Suprises me to see no real usage of this temp file though since that code just writes it to the temp file first and then to the normal one.

GlenDC:
It it surprising indeed.

robvanmieghem:
RemoveFile does not seem to be used either

GlenDC:
I am sure there is more not used. Let's keep that for a new PR where I run some tooling to remove all unused code, that remained in the codebase from past changes by us or Sia.

PR #626, opened 2019-12-17
fix codebase using staticcheck

GlenDC:
Also adds the staticcheck as a travis test,
next to the already ran ineffassign test on Travis.

This is a follow-up of the request on removing unused code in merged PR #624.


modules/blockcreator/update.go
```diff
@@ -9,64 +9,64 @@ import (
 )
 
 // ProcessConsensusChange will update the blockcreator's most recent block.
-func (bc *BlockCreator) ProcessConsensusChange(cc modules.ConsensusChange) {
-	bc.mu.Lock()
-	defer bc.mu.Unlock()
+func (b *BlockCreator) ProcessConsensusChange(cc modules.ConsensusChange) {
```
GlenDC: These kind of fixes are because we were using different local names depending on the method. I simply took the one that was used for the most methods already.

modules/consensus/consensusset.go
```diff
@@ -223,7 +223,7 @@ func New(gateway modules.Gateway, bootstrap bool, persistDir string, bcInfo type
 						cs.log.Println("[ERROR] Failed to close consensus db debug file: ", err)
 					}
 					return
-				case <-time.Tick(time.Second):
+				case <-time.NewTicker(time.Second).C:
```
GlenDC: Global Tick function should only be used for test purposes, as it can Leak the timer.

PR #636, opened 2020-01-13
Add enable/disable minerfees on authcointx extension

DylanVerstraete:


DylanVerstraete:
Don't merge yet, we might need to add an option to by default let all addresses be authorized.

PR #637, opened 2020-01-14
add reverse option to authcointx plugin

DylanVerstraete:
If the `Reverse` bool is provided we reverse the logic of this plugin.

This means that by default all addresses are authorized and can send and receive.


DylanVerstraete:
Closing because of some issues. Investigating now

